PR #17, opened 2026-07-03
fix: deleted duplicated endpoints when specifying --target

lokot0k:
fix: deleted duplicated endpoints when specifiying --target

## What and why

Fixes HTTP proxy routing when --target includes an MCP endpoint path. Previously, a client request to /mcp with --target http://host/mcp was forwarded to /mcp/mcp, causing endpoint-backed streamable HTTP servers to return 404/EOF. The proxy now treats the target path as the exact upstream endpoint and includes a regression test for that case.

## Checklist


kerlenton:
@lokot0k Great fix, thank you! That /mcp/mcp double-path was a sneaky one, and treating the target as the exact endpoint is spot on for MCP. Merging now, thanks for jumping in!

PR #18, opened 2026-07-04
fix: preserve multiline SSE data separators

koriyoshi2041:
## What and why

SSE allows one event to contain multiple `data:` lines. `sseTap` was concatenating those lines without a separator when emitting the observed payload, so a multiline event such as:

```
data: first line
data: second line
```

was observed as `first linesecond line` instead of `first line\nsecond line`.

This keeps the proxied bytes unchanged and only fixes the observed copy by inserting the newline separator between consecutive `data:` fields. It also adds a focused unit test for the multiline case.

## Checklist

- [ ] Updated the README / docs if user-facing behaviour changed

### DIFF
--- a/internal/proxy/http.go
+++ b/internal/proxy/http.go
@@ -200,6 +200,9 @@ func (t *sseTap) line(l []byte) {
 		return
 	}
 	if rest, ok := bytes.CutPrefix(l, []byte("data:")); ok {
+		if t.dataBuf.Len() > 0 {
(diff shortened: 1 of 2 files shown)


kerlenton:
@koriyoshi2041  Nice catch, thank you! Clean fix, and I like that you left the forwarded bytes untouched. You're also the first outside contributor to the project, which honestly made my day.

Issue #22, opened 2026-07-05
Document the post-mortem workflow for reviewing past sessions from logs


kerlenton:
Post-mortem review, opening a session after the fact and looking through it, already works. The shim writes per-session JSONL logs and the TUI backfills past sessions from disk on start. The problem is it is not documented anywhere, so nobody knows it is possible. This came up in the Show HN thread.

So this is mostly a docs task. Write down how to review a past session from its log after it happened. And maybe add a small convenience for opening one specific session, by its id or by a log path.

Good first issue, mostly documentation with maybe a tiny flag.

Raised in the Show HN discussion.

kerlenton:
Done in #24, thanks @kernelpanic888. The post-mortem guide is now in docs/POST_MORTEM.md and linked from the README.

PR #24, opened 2026-07-05
docs: document post-mortem session review from logs

kernelpanic888:
## Summary
- add a post-mortem guide for reviewing captured sessions from JSONL logs
- document the default sessions directory, MCPSNOOP_HOME/XDG_STATE_HOME behavior, and trace-file usage
- link the guide from the README section that mentions backfilled sessions

## Notes
Refs #22. This is docs-only and does not change CLI behavior.

## Test
- git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,8 @@ mcpsnoop
 
 No flags, no socket paths, no startup order to remember. The shim and the UI
 find each other on their own, and the UI backfills past sessions from disk, so
-it doesn't matter whether you open it before or after your client.
+it doesn't matter whether you open it before or after your client. To inspect a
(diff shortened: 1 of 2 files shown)


kerlenton:
@kernelpanic888 Thanks for this, and welcome, this is your first contribution to the project! The guide is clear and accurate, and I can tell you actually read the code, the paths, env vars and keybindings all line up.

I pushed one small tweak to the last part. The line about `--trace-file` being picked up automatically on the next TUI start was not quite right. `--trace-file` writes the log only to that custom path, not to the default sessions directory, so the backfill on a later start does not find it, though a running TUI still sees it live through the socket. I reworded it to match that. Everything else is great. Merging shortly, thank you!

Issue #23, opened 2026-07-05
Add session export, JSON and a self-contained HTML file


kerlenton:
Reading big or deeply nested JSON payloads in the terminal is painful, and there is no easy way to share a captured session with someone or feed it into other tooling. Someone in the Show HN thread asked for a browser view.

A live web UI feels wrong for a terminal tool and is a lot to maintain, so the terminal-native answer is an export command that writes a session to a portable file you can open or share however you want. Other tools do it the same way. tshark picks an output format with a flag, `-T json` being the structured one. mitmproxy saves sessions to HAR and can print to stdout, and exports single flows to curl or httpie. HAR itself is how browsers and Charles and HTTP Toolkit hand off HTTP sessions.

For mcpsnoop the idea is a `mcpsnoop export` subcommand, plus an export action in the TUI, that reads a session from the existing JSONL logs and writes a portable file, so it works post-mortem too. It takes a format flag in the spirit of tshark.

- json gives the parsed frames with the derived data, request and response correlation, durations, status, tool-level isError, not just the raw JSONL. This is the machine-readable one for analysis or benchmarks.
- html gives a single self-contained file with collapsible JSON and in-page search that opens in a browser with no server. This is the readable and shareable one, and the reason a live web UI is not needed.
- text gives an optional pretty-printed plain dump.

It should also be able to write to stdout so it can be piped.

json is close to a quick win since the data already lives in the logs. html is a bit more work because of the template, but still far lighter than a whole web server and frontend.

References
https://www.mitmproxy.org/posts/har-support/
https://www.wireshark.org/docs/man-pages/tshark.html

Raised in the Show HN discussion.

kerlenton:
Done in #26, thanks @lokot0k. `mcpsnoop export -T json|html|text` is now in main, plus an `e` export action in the TUI, and it's documented in the README.

PR #26, opened 2026-07-06
feat: add session export with text/json/html view

lokot0k:
feat: add session export with text/json/html view

## What and why

Adds a `mcpsnoop export` command for turning saved JSONL sessions into portable
artifacts after the fact.

The command supports `-T json|html|text` and can write either to a file or
stdout. JSON exports include the derived session model, not just raw envelopes:
correlated request/response calls, durations, status, MCP tool-level
`isError`, capabilities, and the original frames. HTML exports produce a
self-contained browser-readable file with search and collapsible JSON, so users
can inspect or share a captured session without running a live web UI. Text
exports provide a lightweight pretty dump for terminal use.

The TUI also gets an export action: `e` writes the selected session as HTML, and
`:export json|html|text [path]` allows choosing the format and destination.

The PR solves issue #23 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,18 @@ matches the method, tool, id and payload, while `tool:` `method:` `id:` `kind:`
 to a search tool, and `dir:s2c kind:req` surfaces server-initiated requests
 (sampling, roots). The `?` help lists each token and the values it accepts.
 
+## Exporting sessions
+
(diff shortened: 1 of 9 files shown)


kerlenton:
Really nice work, thank you. The correlated JSON model and the safe HTML embedding are spot on. Works great, merging, this one headlines the next release.

PR #29, opened 2026-07-06
feat: implement open command

### LINKED ISSUE #25: Open a specific session log file in the TUI
Right now the TUI only shows sessions from the default sessions directory, the ones it backfills from disk on start. There is no way to open a session log that lives somewhere else, and that leaves a few real cases uncovered. A log written with `--trace-file` to a custom path, a log a teammate sent you, or a saved test fixture. You captured it, but you cannot review it in the TUI after the fact. This came up while documenting the post-mortem workflow (#22, #24).

The idea is to let mcpsnoop open one specific session file and show it in the TUI. Since the JSONL trace is the same format the TUI already reads on backfill, most of the loading logic is already there, so this is mostly wiring a file path into that same path.

One thing to decide is the CLI shape. A bare positional argument would collide with the shim form `mcpsnoop <server command>`, so a subcommand like `mcpsnoop open ./session.jsonl` or a flag like `mcpsnoop --open ./session.jsonl` is cleaner. The session id and label can come from the file name.

This pairs naturally with the session export work. Capture or export a session to a portable file, share it, and anyone can open it with mcpsnoop and browse it.

elvinka505:
Implements the `mcpsnoop open` command.

- load a saved JSONL session into a fresh Store
- reuse `proxy.Decode` for reading envelopes
- open the session directly in the TUI without starting the hub

Closes #25

### DIFF
--- a/cmd/mcpsnoop/main.go
+++ b/cmd/mcpsnoop/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/kerlenton/mcpsnoop/internal/exporter"
 	"github.com/kerlenton/mcpsnoop/internal/paths"
 	"github.com/kerlenton/mcpsnoop/internal/proxy"
+	"github.com/kerlenton/mcpsnoop/internal/store"
 	"github.com/kerlenton/mcpsnoop/internal/tui"
 )
 
@@ -312,5 +313,43 @@ func runHub() int {
 
 // runOpen opens a persisted JSONL session directly in the TUI.
 func runOpen(args []string) int {
-	panic("TODO")
+	fs := flag.NewFlagSet("mcpsnoop open", flag.ExitOnError)
+
+	fs.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage: mcpsnoop open <session.jsonl>\n\n")
+		fmt.Fprintf(os.Stderr, "Open a persisted session log directly in the TUI.\n")
+	}
+
+	_ = fs.Parse(args)
+
+	if fs.NArg() != 1 {
+		fs.Usage()
+		return 2
+	}
+
+	path := fs.Arg(0)
+
+	f, err := os.Open(path)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "mcpsnoop open:", err)
+		return 1
+	}
+	defer f.Close()
+	st := store.New(0)
+	err = proxy.Decode(f, func(e proxy.Envelope) {
+		st.Ingest(e)
+	})
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "mcpsnoop open:", err)
+		return 1
+	}
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+
+	if err := tui.RunOpen(ctx, st); err != nil {
+		fmt.Fprintln(os.Stderr, "mcpsnoop open:", err)
+		return 1
+	}
+
+	return 0
 }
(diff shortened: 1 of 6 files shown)


cmd/mcpsnoop/main.go
```diff
@@ -305,3 +310,46 @@ func runHub() int {
...
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "mcpsnoop open:", err)
+		return 1
+	}
+	defer f.Close()
+	st := store.New(0)
+	err = proxy.Decode(f, func(e proxy.Envelope) {
+		st.Ingest(e)
+	})
+	if err != nil {
```
kerlenton, line 342: A decode error here is fatal, so a truncated or corrupt last line makes `open` show nothing even though every envelope before it is already in `st`. Logs from a killed or hung run usually end mid line, and that is exactly what a post mortem opens. Better to open the TUI anyway when at least one envelope was read, or treat `io.ErrUnexpectedEOF` as a clean tail. Backfill already tolerates this.

internal/tui/run.go
```diff
@@ -36,3 +36,13 @@ func Run(ctx context.Context, socketPath, sessionsDir string, slow time.Duration
 	}
 	return err
 }
+
+func RunOpen(ctx context.Context, st *store.Store) error {
+	p := tea.NewProgram(New(st), tea.WithAltScreen(), tea.WithContext(ctx))
```
kerlenton, line 42: `New` does no initial `refresh`, so the model only picks up `st` on the first tick at 400ms. Live `Run` gets nudged by the hub sooner, but `open` has nothing to trigger it, so you see an empty table for a beat before the session appears. An initial `refresh` in `New` fixes it and is harmless for live.

internal/proxy/decode.go
```diff
@@ -0,0 +1,21 @@
+package proxy
+
+import (
+	"encoding/json"
+	"io"
+)
+
+func Decode(r io.Reader, emit func(Envelope)) error {
```
kerlenton, line 12: Right home for this, nice. Two small things. Add a doc comment to the exported `Decode`. And a focused test would be good per CONTRIBUTING, give it a couple of envelopes plus a bad line and assert the emit count and the returned error.

internal/tui/run.go
```diff
@@ -36,3 +36,13 @@ func Run(ctx context.Context, socketPath, sessionsDir string, slow time.Duration
 	}
 	return err
 }
+
+func RunOpen(ctx context.Context, st *store.Store) error {
```
kerlenton, line 41: Exported func with no doc comment. A short `// RunOpen ...` line keeps it consistent with the rest of the package.

internal/proxy/decode.go
```diff
@@ -0,0 +1,21 @@
...
+	dec := json.NewDecoder(r)
+
+	for {
+		var env Envelope
+		if err := dec.Decode(&env); err != nil {
+			if err == io.EOF {
```
kerlenton: Optional nit, `errors.Is(err, io.EOF)` is the form the repo leans toward, though `==` is fine for this sentinel.

kerlenton:
Really nice work, this matches the design we talked about and it works well. I left a few inline notes. Sort those out and I will merge it. Thanks for sticking with it.

elvinka505:
Done! Addressed all review comments and pushed the changes. Thanks!

kerlenton:
All addressed, thanks. New does the initial refresh so open paints immediately, and treating unexpected EOF as a clean end is the right call, a truncated tail opens fine now instead of erroring. Doc comments and the decode test are good too. Ran the full gate locally, all green. Approving.

PR #30, opened 2026-07-06
feat: add trace redaction keys

### LINKED ISSUE #20: Add a redact config to scrub secrets from payloads before writing traces
Payload bodies are written to the session logs as they are. So if a secret ends up inside a payload, like an API key or token in a tool argument or in params, it lands on disk in plaintext. A few people in the Show HN thread cared about this.

Worth noting that Authorization and other HTTP headers are already never captured, only the JSON-RPC message bodies are. So this is only about secrets that live inside the payloads themselves.

The idea is a redact step that runs on the observed copy in the trace path and never touches the forwarded bytes. The real client to server stream stays byte for byte the same, only the copy written to the trace gets scrubbed, so the transparency contract holds.

It would be opt-in. You give it a set of JSON key names to scrub, like authorization, token, api_key, password, and it replaces the value with something like [REDACTED]. Optionally it also takes regex patterns for matching values.

Two things are still open. Whether to scrub only the on-disk trace or also the live socket and TUI view. And whether to ship a small default set of key names or keep it fully off by default.

Raised in the Show HN discussion.

leemeo3:
## Summary
- Add opt-in `--redact-key` parsing for stdio and HTTP shim modes.
- Scrub matching JSON object keys recursively before observed trace copies are written or streamed.
- Document key-based trace redaction and its best-effort limits.

## Scope
This redacts trace copies that feed the JSONL log, export, live TUI, and `y` copy path. The proxied client/server bytes remain unchanged, preserving the transparency contract.

## Validation
- `PATH="$PATH:$(go env GOPATH)/bin" make check`
  - `go vet ./...`
  - `staticcheck ./...`
  - `go test -race ./...`

Fixes #20

AI assistance: Implemented with AI assistance from OpenAI Codex; I reviewed the diff and ran the validation above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,8 +64,10 @@ mcpsnoop http --target http://localhost:3000/mcp --listen :7000
 ```
 
 If payloads can contain secrets, opt in to key-based trace redaction. Matching
-JSON fields are scrubbed in saved session logs, while the proxied bytes still
-pass through unchanged.
+JSON fields are scrubbed in observed trace copies, while the proxied bytes still
+pass through unchanged. Redaction is best effort: it only scrubs values under
+matching JSON object keys, so secrets in stderr text, string values under other
+keys, or frames that are not valid JSON pass through.
 
 ```bash
 mcpsnoop --redact-key token,api_key,password -- node build/index.js
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ mcpsnoop http --target http://localhost:3000/mcp --listen :7000
 
 If payloads can contain secrets, opt in to key-based trace redaction. Matching
 JSON fields are scrubbed in observed trace copies, while the proxied bytes still
-pass through unchanged. Redaction is best effort: it only scrubs values under
+pass through unchanged. Redaction is best effort and only scrubs values under
 matching JSON object keys, so secrets in stderr text, string values under other
 keys, or frames that are not valid JSON pass through.
 
--- a/README.md
+++ b/README.md
@@ -63,6 +63,15 @@ For a streamable-HTTP server, run mcpsnoop as a reverse proxy.
 mcpsnoop http --target http://localhost:3000/mcp --listen :7000
 ```
 
+If payloads can contain secrets, opt in to key-based trace redaction. Matching
+JSON fields are scrubbed in saved session logs, while the proxied bytes still
+pass through unchanged.
+
+```bash
+mcpsnoop --redact-key token,api_key,password -- node build/index.js
+mcpsnoop http --target http://localhost:3000/mcp --redact-key authorization
+```
+
 No server of your own? [Try it for real](docs/TRY_IT.md) against a published
 test server, driven by your own client. To inspect a session after it happened,
 see [review past sessions from logs](docs/POST_MORTEM.md).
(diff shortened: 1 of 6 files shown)


cmd/mcpsnoop/main.go
```diff
@@ -244,7 +269,11 @@ func traceSink(sessionID, traceFile string, noTrace bool) proxy.Sink {
 	if f, err := os.OpenFile(traceFile, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0o600); err != nil {
 		fmt.Fprintf(os.Stderr, "mcpsnoop: cannot open trace file %q: %v (continuing without file trace)\n", traceFile, err)
 	} else {
-		sinks = append(sinks, proxy.NewAsyncSink(f, 0))
+		fileSink := proxy.Sink(proxy.NewAsyncSink(f, 0))
+		if redaction.Enabled() {
+			fileSink = proxy.NewRedactingSink(fileSink, redaction)
```
kerlenton: The transparency handling is right, only the sinks are touched and RedactEnvelope allocates a new slice instead of mutating the shared Raw buffer, so the forwarded bytes stay verbatim. The one change I would make is scope. Only the file sink is wrapped, so the socket sink feeds the store unredacted, and the store is what the live TUI and the `y` copy read during a session. `y` is the exact path POST_MORTEM points people to for a bug report, so redaction leaks on the workflow it is meant for, which is worse than not having it since it reads as safe. Wrapping the fan-out covers file and socket in one pass, so the log, export, live view, and copy stay consistent, and the forwarded bytes are still untouched.

    sink := proxy.Sink(proxy.NewMultiSink(sinks...))
    if redaction.Enabled() {
        sink = proxy.NewRedactingSink(sink, redaction)
    }
    return sink

cmd/mcpsnoop/main.go
```diff
@@ -244,7 +269,11 @@ func traceSink(sessionID, traceFile string, noTrace bool) proxy.Sink {
 	if f, err := os.OpenFile(traceFile, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0o600); err != nil {
 		fmt.Fprintf(os.Stderr, "mcpsnoop: cannot open trace file %q: %v (continuing without file trace)\n", traceFile, err)
 	} else {
-		sinks = append(sinks, proxy.NewAsyncSink(f, 0))
+		fileSink := proxy.Sink(proxy.NewAsyncSink(f, 0))
```
kerlenton: Minor, redaction runs synchronously on the pump goroutine, and this AsyncSink Emit right after is the non-blocking handoff it was built for. So a large payload adds decode plus marshal latency to the observe path. Low impact since it is opt in and runs after the frame is already forwarded, but worth knowing.

README.md
```diff
@@ -63,6 +63,15 @@ For a streamable-HTTP server, run mcpsnoop as a reverse proxy.
 mcpsnoop http --target http://localhost:3000/mcp --listen :7000
 ```
 
+If payloads can contain secrets, opt in to key-based trace redaction. Matching
+JSON fields are scrubbed in saved session logs, while the proxied bytes still
```
kerlenton: Worth marking this best effort and naming the edges. It scrubs values under matching JSON object keys only, so secrets in stderr text, in string values not under a matched key, or in a frame that is not valid JSON pass through. A short note keeps users from over trusting it.

PR #31, opened 2026-07-07
feat: add secure live remote capture

### LINKED ISSUE #21: Remote debugging, capture on one host and view on another
Right now the shim and the hub pair over a local socket plus local JSONL logs, so everything sits on one machine. It would be handy to capture MCP traffic on a remote host, where the client or server actually runs, and watch it in a TUI on your own machine. Someone asked for this in the Show HN thread.

Since the traffic already flows to a socket and to JSONL, remote is mostly a transport question. One option is letting the shim stream to a remote sink, something like a tcp address that the hub listens on. Another is leaning on tools people already have, shipping the logs over SSH or tailing them remotely.

The stream carries payloads, so anything going over the network needs auth and encryption. Better to design that in from the start than bolt it on later.

This is a bigger piece, more of a roadmap item.

Raised in the Show HN discussion.

lokot0k:
feat: add secure live remote capture

## What and why

Adds secure remote live capture for MCP traffic.

The TUI hub can now listen for remote shims over TLS, and shims can stream
observed MCP frames to that hub while still writing their local JSONL trace.

Remote capture is opt-in. The default local Unix socket + JSONL workflow is
unchanged.

Previously, live inspection only worked when the MCP client/server and mcpsnoop
TUI were on the same machine. This makes it possible to run the shim where the
MCP traffic actually happens, and watch it live from another machine.

Because captured frames may include prompts, tool args, credentials, and tool
results, the remote stream is TLS-only and requires token auth before the hub
accepts envelopes.

Fixes #21 

## Checklist


kerlenton:
Your write-up is exactly the direction I want, SSH-first with capture staying local and no custom crypto. I am with you on it.

But the PR implements the other option. It adds a hub TLS listener, a shared --remote-token, and --remote-cert / --remote-key / --remote-ca, a custom authenticated transport, and lands the two things your write-up says no to, a new shared token and mcpsnoop-specific TLS cert management.

The SSH-first slice needs almost none of that. The shim already writes the local socket and JSONL, so remote is a documented SSH workflow, forward the hub socket over SSH for live, or copy the remote JSONL and open it locally for post-mortem. Docs plus maybe a small helper, no remotesink, no TLS, no token. Could you repoint the PR at that, keep the README and drop the custom transport? If SSH proves too awkward we can revisit a custom transport later with a proper threat model.

lokot0k:
Agreed. I repointed the PR to the SSH-first slice.

This update removes the custom remote transport entirely: no hub TLS listener,
no shared token, no mcpsnoop-specific cert/key/CA flags, and no `RemoteSink`.
The shim/hub capture model stays local.

The remaining change is documentation for the SSH workflow

kerlenton:
This is exactly right, dropping the custom transport and leaning on SSH. Thanks!!!

PR #40, opened 2026-07-08
feat(cmd): support opening sessions piped directly from stdin

glatinone:
While utilizing the open command to inspect archived sessions, I noticed that it currently requires a physical file path on disk. Supporting direct piping from stdin via the standard dash (-) argument would significantly improve flexibility for automated scripting and quick post-mortem analyses.

Context and impact:
: The open command only accepts an explicit file path string, forcing users to create temporary files on disk if they are fetching session data from external stream sources or remote logs.

Changes introduced:
: Update the file opening logic in cmd/open.go to intercept the dash (-) argument.
: Dynamically assign an io.Reader to utilize either os.Stdin or a secure os.Open stream based on the input argument.
: Pipe the reader straight into the decoder store without altering the terminal UI loading experience.

Functional benefits:
: Enables seamless composition of CLI commands (e.g., cat dump.json | mcpsnoop open -), eliminates the overhead of tracking temporary files on disk, and speeds up diagnostic workflows.

### DIFF
--- a/cmd/mcpsnoop/console_unix.go
+++ b/cmd/mcpsnoop/console_unix.go
@@ -1,0 +1,9 @@
+//go:build !windows
+
+package main
+
+import "os"
+
+func openTTY() (*os.File, error) {
+	return os.OpenFile("/dev/tty", os.O_RDONLY, 0)
+}
--- a/cmd/mcpsnoop/console_windows.go
+++ b/cmd/mcpsnoop/console_windows.go
@@ -1,0 +1,9 @@
+//go:build windows
+
+package main
+
+import "os"
+
+func openTTY() (*os.File, error) {
+	return os.OpenFile("CONIN$", os.O_RDONLY, 0)
+}
--- a/cmd/mcpsnoop/main.go
+++ b/cmd/mcpsnoop/main.go
@@ -13,6 +13,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"io"
 	"net/url"
 	"os"
 	"os/signal"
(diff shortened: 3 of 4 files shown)


kerlenton:
Thanks, and the file path side is clean. There is one significant thing though, and it is partly on the issue for not spelling it out.

When the session comes from stdin, `os.Stdin` is the pipe carrying the data and `proxy.Decode` reads it to EOF. But the TUI reads keyboard input from `os.Stdin` too, since `RunOpen` sets no input and that is bubbletea's default. So after the pipe is drained, `cat file | mcpsnoop open -` launches a TUI whose keyboard is the exhausted pipe. It ends up non-interactive or exits right away on input EOF, which is the whole point of `open`. CI will not catch this because it is a runtime thing, not a compile or test failure.

The fix is to read keys from the controlling terminal for the stdin case instead of `os.Stdin`, and pass it with `tea.WithInput`. The terminal device differs by OS, `/dev/tty` on Unix and `CONIN$` on Windows, so a small build-tagged split handles both. Worth keeping portable since the release ships Windows binaries. It is a small change in `RunOpen` to accept an input reader.

One tiny thing while here, `var f *os.File` can live inside the else since it is only used there.

Sorry the issue did not flag the terminal-versus-stdin part up front. Happy to help with the terminal wiring if you want.

glatinone:
Thank you so much for the incredibly detailed review! That makes total sense regarding how Bubbletea hijacks os.Stdin and leaves the TUI non-interactive after the pipe is drained. I really appreciate you pointing that out.

I have just pushed an update to address this:
- Implemented a cross-platform approach by creating build-tagged console helpers to open the correct controlling terminal device (`/dev/tty` for Unix-like systems and `CONIN$` for Windows).
- Passed the terminal stream securely using `tea.WithInput` so the TUI remains fully interactive when data is piped in.
- Tightened the scope of the `var f *os.File` variable inside the file-handling branch as requested.

Please let me know if this aligns with what you had in mind. Ready for another look!

kerlenton:
Nailed it. Data comes from stdin and keys come from the controlling terminal now, and the build-tagged `openTTY` split keeps it portable, `/dev/tty` on Unix and `CONIN$` on Windows. I checked a Windows cross-build too and it compiles clean. Graceful handling when there is no terminal is a nice touch. Ran the full gate locally, all green. Approving, thanks for the quick turnaround.

PR #54, opened 2026-07-10
feat: add config file support for repeated shim flags

### LINKED ISSUE #37: Support a project config file for default flags
As flags grow (`--label`, `--redact-key`, `--trace-file`, `--no-trace`), repeating them every run gets old, especially `--redact-key` which you want consistent for a project.

Support an optional config file, for example `.mcpsnoop.toml` in the working directory, that sets defaults. Explicit flags override the file. Keep it small to start, just the shim options people repeat.

Worth deciding in the PR. File name and format, search order (cwd only or walk up), and which options are allowed. `MCPSNOOP_HOME` and other env vars should still win where they already do.

elvinka505:
## Summary

Adds optional `.mcpsnoop.toml` support for repeated shim flags.

Supported options:

- `label`
- `trace-file`
- `redact-secrets`
- `redact-key`
- `no-trace`

Precedence is:

```
CLI flags > config file > built-in defaults
```

The config file is looked up only in the current working directory.

## Implementation

- add config loading and parsing
- merge config values after `flag.Parse`
- preserve CLI flag precedence
- add unit tests for parsing and merge behavior
- document the feature in the README

## Notes

This intentionally covers the repeated shim flags discussed in the issue.

Closes #37

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,27 @@ mcpsnoop
 No flags, no socket paths, no startup order to remember. The shim and the UI find
 each other on their own, and the UI backfills past sessions from disk.
 
+### Config file
+
+If you reuse the same shim flags across a project, put them in a
+`.mcpsnoop.toml` file in the current working directory.
+
+```toml
+label = "filesystem"
+trace-file = "trace.jsonl"
+redact-secrets = true
+redact-key = "token,authorization"
+no-trace = false
+```
+
+The config file supports `label`, `trace-file`, `redact-secrets`,
+`redact-key`, and `no-trace`.
+
+The file is only looked up in the current working directory. Parent
+directories are not searched.
+
+Explicit command-line flags override values from the config file.
+
 For a streamable-HTTP server, run mcpsnoop as a reverse proxy.
 
 ```bash
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -11,6 +11,10 @@ mcpsnoop runs the server command you wrap and captures MCP payloads that may
 include credentials, so treat saved traces as sensitive and see the redaction
 options in the README before you share one.
 
+mcpsnoop also reads `.mcpsnoop.toml` from the current working directory. Only
+use it in directories you trust, since a config file can change tracing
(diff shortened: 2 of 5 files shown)


kerlenton:
This is great, good to finally have #37 landed. Precedence is spot on too, `fs.Visit` so the CLI wins, and you even covered `--no-trace=false` beating a config `true`, which is the case people usually miss. make check is green here.

Couple things I'd want sorted before it goes in though.

Main one, it's called `.mcpsnoop.toml` but it's really a small `key = value` reader, not TOML. So things that look like perfectly normal TOML just parse wrong instead of erroring. Ran it on the branch and got this.

- `label = "fs" # my server` ends up with the label `fs" # my server`, comment and quote and all
- `redact-key = ["token", "authorization"]` becomes two keys, `["token"` and `"authorization"]`

The redact-key one is the scary bit, a broken key there means the field you wanted scrubbed just isn't, and nothing warns you. And bools error on a trailing comment while strings don't, so it's inconsistent on top of that. I'd either make the parser strict so that junk errors out (strip a trailing ` #`, reject `[` or `]` with a real message), or drop the `.toml` name so it stops implying full TOML. Just not silently wrong under a `.toml` extension.

Smaller stuff.

- `mcpsnoop http` ignores the file even though it's got the same `label`/`no-trace`/`redact-secrets` flags. Either wire it in there too or just say stdio-only in the readme.
- `loadConfig` runs right after parse, so a broken config in the cwd also kills `mcpsnoop version` and the plain TUI. Would be nicer to only read it when there's an actual `-- server` to run.
- It's read from the cwd, so a `.mcpsnoop.toml` sitting in someone else's repo can point `trace-file` wherever you can write, or flip `no-trace` on. Minor, but worth a line in SECURITY.md so it's written down somewhere.

Solid work, keen to merge once these are in.

elvinka505:
@kerlenton Thanks! I've addressed all the review comments. CI is passing now.

kerlenton:
Nice work, shipping it!

PR #55, opened 2026-07-10
feat: add value redaction patterns

### LINKED ISSUE #46: Redact secret values by pattern, not just by key
Redaction today is key based, so a secret in a string value that is not under a matched key, or in stderr text, passes through. This was noted when the fan out redaction landed.

Add opt in value level scrubbing, for example `--redact-value <regex>` that replaces substrings matching a pattern anywhere in a frame, or a built in heuristic for obvious tokens. It layers onto the existing `RedactConfig` and the same fan out sink, so the saved file and the live stream stay consistent.

Keep it clearly best effort in the docs, since regex over payloads has false positives and misses.

lokot0k:
feat: add value redaction patterns

## What and why

Adds opt-in value-level redaction for captured MCP traces.
This introduces --redact-value <regex> for both stdio and HTTP proxy modes. It layers on top of the existing key-based redaction and uses the same fan-out sink, so saved JSONL traces and the live TUI stream receive the same scrubbed copy while proxied client/server bytes remain unchanged.
The new redaction mode replaces regex matches in observed JSON string values, stderr text, and best-effort non-JSON payload bytes. Docs call out the expected limitations: regex redaction can miss encoded/transformed secrets and can overmatch harmless text.
Tests cover CLI parsing, invalid regex rejection, JSON string redaction, stderr redaction, non-JSON raw redaction, and existing key/preset behavior.

fixes #46 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,11 +236,13 @@ run untrusted ones in a container. It never executes anything you didn't put in
 your client config.
 
 Captured frames can include prompts, tool arguments, credentials, and tool
-results. If payloads can carry secrets, opt in to key-based redaction to scrub
-matching JSON fields from the observed trace copies while the proxied bytes still
-pass through unchanged. It is best effort and only scrubs values under matching
-JSON object keys, so secrets in stderr text, string values under other keys, or
-non-JSON frames pass through.
+results. If payloads can carry secrets, opt in to redaction to scrub the
+observed trace copies while the proxied bytes still pass through unchanged.
+Key-based redaction replaces whole values under matching JSON object keys.
+Value-based redaction applies regular expressions to observed string values,
+stderr text, and non-JSON text frames. Both modes are best effort: regexes can
+miss secrets, overmatch harmless text, or fail to see transformed/encoded
+values.
 
 ```bash
 # built-in preset of common secret keys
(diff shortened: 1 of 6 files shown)


kerlenton:
Nice, this is a genuinely useful layer and the core is right. Redacting by walking the parsed JSON tree so only string values get touched, keys and numbers left alone, is the correct call, and I like that you didn't comma-split the regexes since patterns can have commas in them. Regex is validated at parse time, it's wired into both stdio and http, and Go's RE2 means no catastrophic backtracking to worry about. The JSON and stderr tests look good.

Two things before it can go in.

It's got conflicts with main right now, GitHub shows it as not mergeable, so it needs a rebase. Worth doing that first since it touches the same redaction path #50 changed.

And that rebase matters for a real reason. Your branch already has #50's splitObserved, which means any non-JSON line lands in env.Text now, not env.Raw. So env.Raw is always valid JSON and the "Decode failed" branch in redactRaw never actually runs. Trouble is, if it ever did, it hands back the scrubbed text as a json.RawMessage, and the sink can't encode a RawMessage that isn't valid JSON, so that frame just gets dropped silently, which is the exact trap #50 fixed. TestRedactingSinkAppliesValuePatternToInvalidRaw passes only because it checks the redactor output directly instead of pushing it through a real sink, so it hides that. Non-JSON is already handled by the Text branch you added, so I'd just drop that redactRaw branch and its test and let Text carry it.

Small one, the --redact-value help line only mentions string values and stderr, not the non-JSON bytes, though the readme covers it fine. Not a big deal.

Really solid otherwise, keen to take it once it's rebased and that branch is gone.

lokot0k:
Did as requested

kerlenton:
Perfect, that's everything. Shipping it, thanks for the quick turnaround!

PR #60, opened 2026-07-10
feat: export sessions as OTLP spans

### LINKED ISSUE #49: Export a session as OpenTelemetry spans
Teams running MCP in production lean on OpenTelemetry and tools like Jaeger and Grafana. A captured mcpsnoop session maps cleanly onto spans, one per correlated call, with the method as the name and the duration, status, and tool level isError as attributes.

Add an export format, for example `-T otlp`, that emits OTLP JSON so a session can be dropped into an existing tracing backend. It reuses the same correlated calls the other exporters read, so it slots next to json, html, and text.

lokot0k:
## What and why

Add `mcpsnoop export -T otlp` to export correlated calls as OpenTelemetry spans. Each captured session becomes one trace with a span per correlated MCP call, preserving its method, duration, status, and tool-error context. The OTLP JSON can be imported into existing tracing backends such as Jaeger and Grafana. Closes #49.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ dir:s2c kind:req                  # server-initiated requests (sampling, roots)
 Turn any captured session into a portable file.
 
 ```bash
-mcpsnoop export -T json|html|text [-o file|-] [session-id|log.jsonl]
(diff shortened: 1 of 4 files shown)


kerlenton:
This is great, the OTLP output is spot on. Thanks!

PR #69, opened 2026-07-10
feat: show stream footer signal counts

### LINKED ISSUE #68: Show per-signal counts in the stream footer
The live footer shows total frames and an error count. Now that the stream flags several distinct problems (err, slow, bad for invalid frames, warn for validation issues), showing each count would let you gauge a session's health at a glance without scrolling, for example `142 frames · 3 err · 2 bad · 1 warn`. Small, self-contained change in the TUI footer rendering (internal/tui) over counts the store tracks or a quick pass of the timeline. Good entry point for a first contribution.

leemeo3:
## What and why

Fixes #68.

The stream footer now summarizes each non-zero signal type it can show in the frame table: `err`, `bad`, `warn`, and `slow`. This keeps the existing frame count while making session health visible without scrolling through the stream.

The counts are derived in `internal/tui` from the rendered timeline, so the change stays scoped to the TUI footer and does not alter capture, storage, export, or proxy behavior.

## Checklist

- [x] Updated the README / docs if user-facing behaviour changed (N/A: compact footer status only)

Validation:

```sh
go test ./internal/tui
PATH="$PATH:$(go env GOPATH)/bin" make check
```

AI assistance was used to prepare this contribution.

### DIFF
--- a/internal/tui/view.go
+++ b/internal/tui/view.go
@@ -89,7 +89,21 @@ func (m Model) renderStatus() string {
 		count = fmt.Sprintf("%d/%d frames", len(m.timeline), m.total)
 	}
 	left := live + sep + m.styles.dim.Render(count)
-	if e := m.totalErrors(); e > 0 {
+	if m.view == viewStream {
+		c := m.streamSignals
+		if c.errors > 0 {
+			left += sep + m.styles.respErr.Render(fmt.Sprintf("%d err", c.errors))
+		}
+		if c.bad > 0 {
+			left += sep + m.styles.invalid.Render(fmt.Sprintf("%d bad", c.bad))
+		}
+		if c.warn > 0 {
+			left += sep + m.styles.warn.Render(fmt.Sprintf("%d warn", c.warn))
+		}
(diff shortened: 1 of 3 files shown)


internal/tui/view.go
```diff
@@ -89,7 +89,21 @@ func (m Model) renderStatus() string {
 		count = fmt.Sprintf("%d/%d frames", len(m.timeline), m.total)
 	}
 	left := live + sep + m.styles.dim.Render(count)
-	if e := m.totalErrors(); e > 0 {
+	if m.view == viewStream {
+		c := countStreamSignals(m.timeline, m.store.SlowThreshold())
```
kerlenton: Move this off the render path. countStreamSignals scans the whole timeline, and renderStatus runs on every View, so every scroll and resize repeats a full O(n) pass on a large session. The counts only change when the timeline is rebuilt at model.go:603 (m.timeline = m.filterEvents(full)), so compute them once there and store the result on the model, then renderStatus just reads the cached struct. Same output, no per-keystroke rescan.

PR #74, opened 2026-07-11
feat: show used and unused tools in capabilities view

### LINKED ISSUE #66: Show which advertised tools were never called
A common failure mode is the model silently never calling a tool it should have. mcpsnoop already sees the server's tools/list (advertised tools) and every tools/call. Surface the gap: which advertised tools were actually invoked in a session and which were never touched. Could live in the capabilities inspector (press c) as a used and unused list, or as a line in the session summary. Makes "the tool the model never called" visible instead of something you infer from staring at the stream.

elvinka505:
## Summary

Display advertised MCP tools in the capabilities view, grouped into:

- USED
- UNUSED

This works even when the initialize handshake was not captured but a `tools/list` response was observed.

## Changes

- track tools advertised via `tools/list`
- expose tool usage through `ToolUsage`
- show USED/UNUSED sections in the capabilities panel
- add store tests covering used/unused classification and multiple `tools/list` responses

Closes #66


internal/store/views.go
```diff
@@ -172,6 +172,35 @@ func (s *Store) Capabilities(sessionID string) (CapsView, bool) {
...
+
+	called := make(map[string]bool)
+
+	for _, c := range sess.calls {
+		if c.isTool && c.toolName != "" {
+			called[c.toolName] = true
+		}
+	}
+
+	for _, name := range sess.advertisedTools {
```
kerlenton, line 193: This walks only advertisedTools, so a tool that was called but never advertised ends up in neither USED nor UNUSED and just disappears. For a usage view that is a real mismatch worth showing (a stale advertised list, the wrong server, or a missed tools/list page). You already build the full called set right above, so also report called minus advertised, either as a third group or folded into USED with a marker. I know I called this optional earlier, but in context I think it should go in.

internal/tui/view.go
```diff
@@ -585,15 +585,27 @@ func (m Model) capsContent() string {
...
-	b.WriteString(m.styles.resp.Render(" SERVER  "+infoLine(caps.ServerInfo)) + "\n")
-	b.WriteString(capsBlock(caps.Server))
+	if ok {
+		b.WriteString(" protocolVersion: " + valueOr(caps.ProtocolVersion, "(unknown)") + "\n\n")
+		b.WriteString(m.styles.req.Render(" CLIENT  "+infoLine(caps.ClientInfo)) + "\n")
+		b.WriteString(capsBlock(caps.Client) + "\n\n")
+		b.WriteString(m.styles.resp.Render(" SERVER  "+infoLine(caps.ServerInfo)) + "\n")
+		b.WriteString(capsBlock(caps.Server))
+	} else {
+		b.WriteString(m.styles.dim.Render(" handshake not captured"))
```
kerlenton: Same string as the early return at line 591. The else here only runs when there are tools to show, so hoist the no-caps message to one place so the two copies cannot drift.

elvinka505:
@kerlenton Thanks! Addressed both comments. Added the CALLED BUT NOT ADVERTISED section, hoisted the repeated string into a local variable, and added a test (covering multiple unadvertised tools with stable ordering) 🙂

kerlenton:
Thanks! LGTM!

PR #78, opened 2026-07-11
feat: narrow trace redaction to selected JSON values

### LINKED ISSUE #67: Redact values by JSONPath, not just by key or regex
mcpsnoop can redact by JSON key (--redact-key) and by regex over string values (--redact-value). Both are coarse: key redaction hits every field with that name anywhere, and regex can overmatch. A path-scoped option like `--redact-path '$.params.arguments.password'` would scrub exactly one location, which matters when a secret sits under a common key name you do not want to blanket-redact. Layers onto the existing redactor (internal/proxy/redact.go) next to keys and value patterns, using the same non-mutating fan-out so proxied bytes stay untouched.

ychampion:
## What and why

Adds repeatable `--redact-path` support for scrubbing only JSON values selected by a JSONPath expression. This avoids blanket-redacting common keys such as `password` when only one location is sensitive.

Paths are validated once at CLI or config parsing, then applied to the decoded trace copy before the existing key and regex layers. Exact paths, wildcard matches, invalid paths, config precedence, unchanged no-match payloads, and durable/live trace fan-out are covered.

Closes #67.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ label = "filesystem"
 trace-file = "trace.jsonl"
 redact-secrets = true
 redact-key = "token,authorization"
+redact-path = "$.params.arguments.password"
 no-trace = false
 ```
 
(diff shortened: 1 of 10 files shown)


kerlenton:
Thanks! LGTM!

PR #80, opened 2026-07-11
feat: stream completed calls to OTLP collectors

### LINKED ISSUE #64: Stream traces to a live OTLP endpoint, not just a file
#60 added `export -T otlp` to a file. Production observability setups want spans delivered live to a collector (Jaeger, Grafana Tempo, an otel-collector) over OTLP/HTTP so mcpsnoop can feed end-to-end tracing while traffic happens, not after. Add an option to POST the same payload writeOTLP builds to an OTLP/HTTP traces endpoint, for example `--otlp-endpoint http://localhost:4318/v1/traces`, either from `export` or as a live sink alongside the existing file and socket fan-out. Keep it dependency-light, a plain HTTP POST of the JSON, no heavy otel SDK. Closes the loop between mcpsnoop and the tracing backends teams already run.

ychampion:
## What and why

Adds optional live OTLP/HTTP JSON delivery for completed MCP calls in both stdio and HTTP proxy modes. The sink retries collector failures in a background worker, bounds queued and pending frames, drops instead of blocking proxied traffic, and reuses the existing OTLP exporter. Repeatable headers support collector authentication and tenancy.

Closes #64.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,28 @@ Omit `-o` to write to stdout, and omit the session to take the newest, or pass
 `-` to read JSONL from stdin. In the TUI, press `e` to export the selected
 session as HTML, or run `:export json|html|text|otlp [path]` from command mode.
 
+### Stream completed calls to an OTLP collector
+
+Send spans while the proxy is running by pointing it at an OTLP/HTTP JSON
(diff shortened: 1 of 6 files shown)


kerlenton:
LGTM. Thanks!

ychampion:
Thanks!

PR #81, opened 2026-07-12
feat: summarize per-tool session performance

### LINKED ISSUE #45: Session summary with per-tool latency and error stats
mcpsnoop already correlates calls and computes durations, but there is no way to see the shape of a session at a glance. People profiling MCP servers care about per-tool latency and error rates, with common targets like p50, p95, and p99.

Add a summary view, reachable with a key from the sessions or stream table, that aggregates per tool: call count, error count, latency percentiles, and the slowest calls. It reads straight from the correlated calls the store already holds.

This would pair well with export too, so a json export could carry the same summary.

ychampion:
## What and why

Adds an `s` summary view for a selected session with per-tool call, error, and pending counts; p50, p95, and p99 latency; and the five slowest completed calls. Pending calls stay visible in counts without affecting latency percentiles.

JSON exports carry the same store-backed summary, so interactive and automated analysis use identical aggregation semantics.

Closes #45.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,10 +166,11 @@ server is written in.
 | `esc` | back | | `:` | command |
 | `j` / `k` | move | | `r` | replay a call |
 | `g` / `G` | top / bottom | | `c` | capabilities |
-| `ctrl-f` / `ctrl-b` | page | | `y` | copy |
+| `ctrl-f` / `ctrl-b` | page | | `s` | tool summary |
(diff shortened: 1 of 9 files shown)


kerlenton:
Looks good to me, thanks!

ychampion:
Thanks for the review and merge.

PR #103, opened 2026-07-18
feat: detect MCP tool definition drift with TOFU baselines

### LINKED ISSUE #94: Trust-on-first-use baseline: detect tool definition drift (rug pulls) live and in CI
## Context

A **rug pull** is when a server silently changes a tool's definition after the client already approved it, turning a trusted tool into a compromised one with no re-approval. MCP has no built-in mechanism to detect this, and static scanners can't catch it because it only shows up at runtime. The control that actually stops it (Trail of Bits) is **trust-on-first-use (TOFU)**: capture the definition the first time it's seen, and flag if it ever changes.

- Tool poisoning / rug pull background: https://invariantlabs.ai/blog/mcp-security-notification-tool-poisoning-attacks

mcpsnoop is in the data path and already has most of the machinery — this issue promotes it from a manual, two-session comparison into automatic detection.

## Current state

- `internal/sessiondiff/diff.go` already computes `AddedTools`, `RemovedTools`, `ChangedSchemas`, and `HasRegression()` — but only when a user manually runs `diff` between two sessions, and it compares **schemas** only.
- `internal/store/store.go` records advertised tools via `applyToolsList` and tracks `advertisedTools` per session.
- `internal/tui/view.go` already has a narrow "DRIFT" notion (tools the client called that the server never advertised) — different from definition drift.

## Proposed change

1. Persist a per-server **baseline** of tool definitions on first sight, stored next to sessions (see `docs/POST_MORTEM.md` for the state dir: `~/.local/state/mcpsnoop/sessions/…`, honoring `MCPSNOOP_HOME` / `XDG_STATE_HOME`).
2. On every `tools/list`, diff the current definitions against the baseline automatically and surface drift in the TUI (a clear banner/marker, not buried).
3. **Also capture and diff the tool `description`, not just the schema** — rug pulls and tool poisoning hide in descriptions. This is the key gap versus the current `sessiondiff`.
4. Expose the result so `check` can gate on it (see #6, `--no-drift`).

## Scope boundary

Detection and highlighting only. mcpsnoop reports drift; it does not block calls or make trust decisions — that stays with the user and with dedicated tooling. This keeps the "Wireshark for MCP" identity intact.

## Acceptance criteria

- [ ] First-seen tool definitions are persisted as a baseline per server
- [ ] Description **and** schema changes against the baseline are detected
- [ ] Drift is surfaced prominently in the TUI on the next `tools/list`
- [ ] A command/flag prints drift for a captured session (feeds #6)
- [ ] Baseline location respects `MCPSNOOP_HOME` / `XDG_STATE_HOME`
- [ ] There's a way to accept/reset the baseline after a legitimate change
- [ ] Tests cover: unchanged, description-only change, schema change, added/removed tool

## Files

- `internal/sessiondiff/diff.go` (extend to descriptions; reuse `Compare`)
- `internal/store/store.go` (`applyToolsList`, capture descriptions)
- `internal/paths/paths.go` (baseline path)
- `internal/tui/view.go` (surface drift)

jstar0:
## Summary

Persist a trust-on-first-use baseline for each server label and flag later MCP tool-definition changes in the TUI and CI.

The baseline includes tool descriptions and input schemas, so changes that can alter model behavior are visible even when tool names stay the same. Added and removed tools are reported as drift as well.

## Changes

- retain complete paginated `tools/list` definitions in the store
- report description changes from the existing captured-session `diff` path
- persist versioned baselines under the existing mcpsnoop state directory
- detect description, schema, added-tool, and removed-tool drift
- show drift in the sessions table and tool summary
- add `mcpsnoop baseline` to inspect, accept, or reset trusted definitions
- add opt-in `drift` support to `mcpsnoop check --fail-on`

## Scope and risk

This is detection and reporting only. It does not block calls, change trust automatically, or touch the proxy byte path.

Baselines are keyed by the existing server label, so users with otherwise-colliding command names or target hosts should set a stable unique `--label`. The default `check` signals remain unchanged.

Concurrent first observations use exclusive creation so one process cannot overwrite another process's first trusted definition set. Invalid baseline state is surfaced as an error rather than silently ignored.

## Verification

```bash
make check
go test -race ./internal/toolbaseline -run TestConcurrentFirstObservationDoesNotOverwriteTrustedDefinition -count=100
GOOS=windows GOARCH=amd64 make build
GOOS=linux GOARCH=amd64 make build
```

Closes #94

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ Explicit command-line flags override values from the config file.
 | `mcpsnoop http --target <url>` | proxy a streamable-HTTP server |
 | `mcpsnoop export` | render a session to json, html, text, or otlp |
 | `mcpsnoop check` | fail CI on errors, invalid frames, warnings, routing mismatches, or hung calls |
+| `mcpsnoop baseline` | inspect, accept, or reset trusted tool definitions |
 | `mcpsnoop diff` | compare tools and calls across two captured sessions |
 | `mcpsnoop open` | open a saved session in the TUI |
 | `mcpsnoop remote <user@host>` | print the SSH tunnel command |
@@ -112,6 +113,7 @@ Run `mcpsnoop help` for the full list, or `mcpsnoop help <command>` for the flag
 | Flags hung calls and stream errors | no | yes |
 | Flags stray output that corrupts the stream | no | yes |
 | Flags malformed JSON-RPC frames | no | yes |
+| Detects tool definition drift after approval | no | yes |
 | Interactive terminal UI | no | yes |
 | Zero-config, no flags or ordering | no | yes |
 | Capability inspector | partial | yes |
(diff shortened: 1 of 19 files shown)


kerlenton:
LGTM! Thanks!

Issue #107, opened 2026-07-18
Bound the hub backfill and add session log retention


kerlenton:
## Context

Nothing in the repository prunes anything. Every shim run appends a JSONL trace to the sessions directory, and nothing ever removes it except a manual delete in the TUI. On every start the hub replays every `.jsonl` it finds into the store, and the store keeps all of it in memory for the life of the process.

The result compounds over time. The sessions directory grows for as long as mcpsnoop is used, TUI startup gets slower in proportion to the entire history rather than to what the user cares about, resident memory tracks the same total, and `hub.seen` gains an entry per session that is never released.

For a tool that lives in the background of a developer's day this shows up eventually on every machine, not only in unusual setups.

## Current state

- `cmd/mcpsnoop/main.go` points the shim's trace file at `paths.SessionLogPath(sessionID)` when no explicit `--trace` is given.
- `internal/hub/hub.go`, `backfill` lists every `.jsonl` in the sessions directory, sorts by name, and calls `replayFile` on all of them before accepting connections.
- `internal/hub/hub.go`, `emit` records a high-water mark per session in `h.seen`, which is never pruned.
- `internal/store` holds every event of every session for the process lifetime.
- The only deletion path is `deleteCurrentSession` in `internal/tui/model.go`, one session at a time, driven by the user.

## Two separable pieces

These are worth keeping apart, because one is safe and one destroys user data.

**Bounded backfill is safe and should come first.** Replaying only the most recent slice of history fixes startup time and memory without touching a single file. Older traces stay on disk and remain reachable through `mcpsnoop open <session>` and `mcpsnoop export`, which resolve a path directly rather than going through the hub.

**Retention deletes traces, so it should be opt-in and explicit.** A trace is evidence. Silently discarding it because it aged out is exactly the wrong behavior for a debugging tool, so a default-on policy needs a strong argument.

## Proposed change

1. Bound what `backfill` replays. Pick a limit that is easy to reason about, for example the newest N sessions or everything newer than a given age, with a flag to override and a way to ask for everything. Say in the TUI when history was truncated, so a missing old session looks deliberate rather than lost.
2. Release `h.seen` entries for sessions the hub no longer tracks, or bound the map the same way the backfill is bounded.
3. Add an explicit way to prune the directory, for example `mcpsnoop prune --older-than 30d` with a dry run. Keep it manual, or opt-in through config, rather than silently on by default.

## Acceptance criteria

[...]

PR #110, opened 2026-07-18
feat: bound saved session backfill

floze-the-genius:
## What and why

Implements the safe, non-destructive first part of #107. The live hub now replays only the newest 100 saved sessions by default, which bounds startup I/O, store growth from historical data, and the initial `seen` map population.

`--history-limit N` overrides the bound and `--history-limit 0` restores full-history loading. When history is truncated, the TUI reports how many sessions were loaded and makes clear that older traces remain on disk. Those traces still resolve directly through `mcpsnoop open <session-id>` and `mcpsnoop export <session-id>`.

This intentionally does not add automatic retention or deletion. A prune workflow is destructive and can follow separately with explicit dry-run semantics.

## Tests

- `make check` (gofmt, go vet, staticcheck, full `go test -race ./...`)
- `go build ./...`
- regression coverage for newest-N selection, backfill reporting, preservation of out-of-bound logs, CLI flag wiring, and the visible truncation notice

## Checklist

- [x] Added or updated tests for the change
- [x] Updated the README for user-facing behavior

Implementation and tests were prepared with AI assistance and manually validated against the repository contribution guide and full local gate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,12 @@ shim your client spawns, forwarding bytes verbatim while shipping a copy of ever
 frame to the hub. `mcpsnoop` with no arguments is that hub and its live TUI. They
 pair through a well-known socket and on-disk logs, so neither has to start first.
 
+The hub loads the newest 100 saved sessions by default, keeping startup work
+bounded without deleting older traces. Use `mcpsnoop --history-limit N` to pick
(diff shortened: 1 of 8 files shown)


kerlenton:
LGTM! Thanks!

PR #124, opened 2026-07-19
feat(store): flag deprecated Roots, Sampling, and Logging usage

### LINKED ISSUE #122: Flag use of Roots, Sampling and Logging, deprecated in 2026-07-28
## Context

The 2026-07-28 release deprecates Roots, Sampling and Logging under the new feature lifecycle policy. The deprecation is annotation only, the methods and capability flags keep working in this release and in every version published within a year of it, and removal needs a separate SEP. So nothing breaks today, but everyone using them is on a clock with a defined migration path.

- Release candidate: https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/

The replacements are specific. Roots gives way to tool parameters, resource URIs or server configuration. Sampling gives way to direct integration with an LLM provider API. Logging gives way to stderr for stdio and OpenTelemetry for structured observability, both of which mcpsnoop already handles.

Telling someone their server still speaks a deprecated method is exactly the sort of thing that is invisible until a migration deadline arrives, and mcpsnoop is already watching every frame.

## Current state

Nothing in the codebase mentions the deprecation. `validationWarning` in `internal/store/store.go` is where a per-frame protocol warning is attached today.

There is a neat coincidence in `internal/tui/view.go`. The capability inspector already lists these exact features, `clientCapOrder` is `roots, sampling, elicitation` and `serverCapOrder` is `tools, resources, prompts, logging, completions`, so three of the entries it already renders are the three now deprecated.

## Proposed change

1. Flag a frame whose method belongs to a deprecated feature, that is the `roots/`, `sampling/` and `logging/` families, including their notifications. Name the replacement in the message, since the whole value is telling someone what to do instead.
2. Mark the same three entries in the capability inspector, so a server advertising a deprecated capability reads as deprecated at a glance rather than only when it happens to use it.

Reuse whatever mechanism carries per-frame markers today rather than inventing a new one, and follow how the truncation marker was separated from protocol warnings, a deprecation is a heads-up rather than a protocol violation and should not fail a default `check --fail-on warn` on its own.

## Scope boundary

Flagging only. Do not block, rewrite or warn on the wire, and do not add a `check` signal for it in this change. Whether a CI gate should be able to fail on deprecated usage is a reasonable follow-up, but it needs its own decision.

## Acceptance criteria

- [ ] A `roots/`, `sampling/` or `logging/` frame is flagged, with the replacement named
- [ ] The three capabilities are marked as deprecated in the capability inspector
- [ ] A default `check` run does not fail because a session used a deprecated feature
- [ ] Non-deprecated methods are untouched, no false positives on a method that merely starts with a similar string
- [ ] Tests cover one method from each family, a notification, and a negative case

## Files

- `internal/store/store.go`, near `validationWarning`
- `internal/tui/view.go`, `clientCapOrder`, `serverCapOrder`, `capSection`

syf2211:
## Summary

Flag MCP frames and capabilities that use features deprecated in the 2026-07-28 release (SEP-2577): Roots, Sampling, and Logging.

## Motivation

Developers debugging MCP sessions need to know when traffic still relies on deprecated features before the removal window closes. mcpsnoop already surfaces protocol warnings and capability declarations; this adds deprecation heads-ups in the same places without treating them as check failures.

## Changes

- Add `deprecatedMethodNote()` and `DeprecatedCapabilityNote()` in `internal/store/deprecation.go`
- Set a structured `Deprecated` field on timeline events (parallel to `Truncated`, not routed through `Warning`)
- Show deprecation notes in the TUI stream, `status:warn` filter, and capability inspector for `roots`, `sampling`, and `logging`
- Export the `deprecated` field in JSON, text, and HTML exporters
- Add unit tests for method detection, ingest behavior, and default `check` pass-through

## Tests

```bash
make check
```

- `go vet`, `staticcheck`: pass
- `go test -race ./...`: pass for all packages except two pre-existing TUI style tests on `main` (`TestStatusStyleWarnsOnTruncated`, `TestStreamRowShowsSupersededStatusInWarnStyle`) that fail in this environment
- New tests: `TestDeprecatedMethodNote`, `TestIngestDeprecatedMethods`, `TestIngestDeprecatedMethodNegative`, `TestCheckPassesForDeprecatedFeature`

## Notes

- Deprecation is detected from the frame method at ingest time; responses are not flagged (method is empty).
- Default `mcpsnoop check --fail-on warn` does not fail on deprecated usage, matching the issue scope boundary.
- `notifications/roots/*` and `notifications/message` are covered alongside the `roots/`, `sampling/`, and `logging/` method families.

Fixes #122

### DIFF
--- a/cmd/mcpsnoop/check_test.go
+++ b/cmd/mcpsnoop/check_test.go
@@ -469,6 +469,19 @@ func TestCheckPassesForTruncatedBody(t *testing.T) {
 	}
 }
 
+func TestCheckPassesForDeprecatedFeature(t *testing.T) {
+	t.Setenv("MCPSNOOP_HOME", t.TempDir())
+	deprecated := checkEnvelope(1, proxy.ClientToServer, `{"jsonrpc":"2.0","id":1,"method":"roots/list"}`)
+
+	code, stdout, stderr := executeCheck(t, []string{"-"}, encodeCheckLog(t, deprecated))
+	if code != 0 || stderr != "" {
+		t.Fatalf("a deprecated feature must not fail the default check, code %d stderr %q", code, stderr)
+	}
+	if !strings.Contains(stdout, "warnings=0") || !strings.Contains(stdout, "check passed") {
+		t.Fatalf("stdout = %q", stdout)
+	}
+}
+
 func executeCheck(t *testing.T, args []string, stdin string) (int, string, string) {
 	t.Helper()
 	cmd := newCheckCmd()
--- a/internal/exporter/exporter.go
+++ b/internal/exporter/exporter.go
@@ -105,18 +105,19 @@ type CallExport struct {
 }
 
 type EventExport struct {
-	Seq       uint64          `json:"seq"`
-	Timestamp time.Time       `json:"timestamp"`
-	Direction proxy.Direction `json:"direction"`
-	Kind      string          `json:"kind"`
-	Method    string          `json:"method,omitempty"`
-	ID        string          `json:"id,omitempty"`
-	Warning   string          `json:"warning,omitempty"`
-	Mismatch  bool            `json:"mismatch,omitempty"`
-	Truncated bool            `json:"truncated,omitempty"`
-	CallIndex *int            `json:"call_index,omitempty"`
-	Raw       json.RawMessage `json:"raw,omitempty"`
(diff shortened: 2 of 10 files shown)


kerlenton:
Looks good, merging. Thanks!

PR #125, opened 2026-07-20
feat(store): correlate task-backed tool calls

### LINKED ISSUE #121: Correlate task-backed tool calls, or every long-running call reads as instant
## Context

The 2026-07-28 release ships Tasks as an official extension (`io.modelcontextprotocol/tasks`, SEP-2663). A server can answer a `tools/call` with a task handle instead of a result, and the client then drives the work with `tasks/get`, `tasks/update` and `tasks/cancel`.

- Release candidate: https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/
- Extension docs: https://modelcontextprotocol.io/extensions/tasks/overview

This breaks the one thing mcpsnoop exists to get right. `completeCall` in `internal/store/store.go` matches a response to its request purely by JSON-RPC id, so a task-backed call is marked complete the moment the handle comes back. Against a server using Tasks, mcpsnoop reports a call that ran for ten minutes as having taken five milliseconds, never flags it as hung when it stalls, and shows the polling as a wall of unrelated `tasks/get` frames with no link to the call they belong to. Latency stats, the pending signal and `check --max-duration` are all wrong in the same direction, and nothing says so.

Nothing in the codebase mentions `tasks/` today.

## The shape to follow

Taken from the extension docs, so the field names below are the real ones.

- The server responds to a supported request with a `CreateTaskResult`, identified by `resultType: "task"`, carrying a `taskId`, an initial `status`, a TTL and a `pollIntervalMs`.
- The client polls `tasks/get` with the `taskId` until a terminal status.
- Statuses run `working`, `input_required`, then one of the terminal `completed`, `failed`, `cancelled`.
- On `completed` the `result` field holds what the original request would have returned synchronously. On `failed` the `error` field holds the JSON-RPC error.
- `input_required` means the `tasks/get` response carries an `inputRequests` map, which the client answers with `tasks/update`, and the task returns to `working`.
- `tasks/cancel` is acknowledged with an empty result and is cooperative, so the task may still end in a non-cancelled terminal status.
- A server may also push `notifications/tasks`, each carrying the same full task state a `tasks/get` response would.
- There is no `tasks/list`, so a task can only ever be reached through its id.

## Proposed change

Treat a task handle as the continuation of the originating call rather than as its answer.

1. When a `tools/call` response carries `resultType: "task"`, keep the call open, record its `taskId`, and mark it as task-backed rather than completed.
2. Link every later `tasks/get`, `tasks/update`, `tasks/cancel` and `notifications/tasks` frame carrying that `taskId` back to the originating call.
3. Close the call on a terminal status, taking the outcome from `result` or `error`, so its duration spans the real work and a failed task reads as a failed call.
4. Keep it pending until then, so the pending signal, `check --fail-on pending` and `check --max-duration` all judge the work rather than the handshake.
5. Show the link in the stream, so the polls read as belonging to their call instead of as unrelated traffic. A filter token for the task id would help here, matching how `id:` already works.

## Scope boundary

Observation only. mcpsnoop never creates, polls or cancels a task, and never subscribes to `notifications/tasks` on its own, it only reads what the real client and server exchange. Leave the separate question of threading calls by application state handle alone, that is a different problem with a different solution.

## Acceptance criteria

- [ ] A `tools/call` answered with `resultType: "task"` stays open rather than completing
- [ ] Later `tasks/*` frames and `notifications/tasks` for the same `taskId` are linked to the originating call
- [ ] The call closes on a terminal status, taking its result or error from the terminal payload
- [ ] Its duration spans request to terminal status, so latency stats and `--max-duration` measure the real work
- [ ] A task still `working` counts as pending
- [ ] An `input_required` task is visibly waiting on the client rather than looking stalled on the server
- [ ] A cancelled task that still reaches a different terminal status is reported by its actual outcome
- [ ] A `tasks/get` whose originating call was never observed degrades gracefully, no crash and no invented parent
- [ ] Tests cover the full lifecycle, an input-required round trip, a failed task, a cancel, and an orphan `tasks/get`

## Files

- `internal/store/store.go`, `call`, `callKey`, `openCall`, `completeCall`, `Ingest`
- `internal/store/views.go`, `CallView` and the pending accounting
- `internal/tui/view.go` and `internal/tui/model.go` for the link and the filter token

w3lld1:
## What and why

I changed task-backed `tools/call` correlation so a task handle keeps the original call open until the observed task reaches a terminal state. Task lifecycle frames now link back to that call, terminal results and errors drive its outcome, and duration/pending accounting cover the full task lifetime.

The stream shows the task-to-call link, highlights `input_required`, and supports a `task:` filter. I also added lifecycle coverage for completion, failure, cooperative cancellation, input updates, notifications, and orphan polls.

Fixes #121

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,7 @@ matches the method, tool, id, and payload.
 | `tool:` | tool name | `tool:search` |
 | `method:` | JSON-RPC method | `method:tools/call` |
 | `id:` | request id | `id:7` |
+| `task:` | task id | `task:01J...` |
 | `dir:` | direction (`c2s`, `s2c`) | `dir:s2c` |
 | `kind:` | frame type (`req`, `resp`, `notify`, `stderr`, `invalid`) | `kind:invalid` |
 | `status:` | call outcome (`ok`, `error`, `pending`, `bad`, `warn`, `mismatch`) | `status:error` |
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -74,18 +74,20 @@ type callKey struct {
 
 // call is the mutable internal record for one request/response pair.
 type call struct {
-	id       string
-	method   string
-	reqDir   proxy.Direction
-	params   json.RawMessage
-	result   json.RawMessage
-	err      *proxy.RPCError
-	start    time.Time
-	end      time.Time
-	state    CallState
-	isTool   bool
-	toolName string
-	toolErr  bool // result.isError == true (MCP tool-level failure)
+	id         string
+	method     string
+	reqDir     proxy.Direction
+	params     json.RawMessage
+	result     json.RawMessage
+	err        *proxy.RPCError
+	start      time.Time
+	end        time.Time
+	state      CallState
+	isTool     bool
+	toolName   string
+	toolErr    bool // result.isError == true (MCP tool-level failure)
+	taskID     string
+	taskStatus string
 }
 
 // event is the mutable internal timeline entry.
@@ -107,6 +109,8 @@ type event struct {
 	truncated          bool   // the observed copy was capped at the frame-size limit
 	deprecated         string // a deprecated MCP feature was used (structured, not a protocol warning)
 	call               *call  // set for request/response events
+	taskCall           *call  // originating call for a task lifecycle frame
+	taskID             string
 }
 
 // capabilities holds what each side declared, whether through the legacy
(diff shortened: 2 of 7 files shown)


kerlenton:
Looks good! Thanks!

PR #141, opened 2026-07-22
fix(check): summarize max-duration failures

### LINKED ISSUE #135: --max-duration prints one line per slow call and floods the CI log
## Context

`checkAssertions.eval` appends a message for every tool call over the budget, so a session with a hundred slow calls emits a hundred lines. The signal that matters is that the budget was breached and by how much at worst, and the volume buries it in exactly the place people read least carefully.

## Proposed change

Collapse the latency budget into one message per session naming how many calls exceeded it and the worst offender, or cap the detail at the first few and say how many more there were. `--expect-tool` and `--forbid-tool` already emit one line each and should stay as they are.

## Acceptance criteria

- [ ] A session with many slow calls produces a bounded number of lines
- [ ] The message names the count and the worst case
- [ ] The exit code is unchanged
- [ ] The other two assertions are untouched
- [ ] Test with several calls over budget

## Files

- `cmd/mcpsnoop/check.go`

edwei06:
## What and why

Collapse all `--max-duration` violations for a session into one assertion message containing the breach count and worst call. This keeps CI output bounded for large captures while preserving the non-zero exit behavior and leaving `--expect-tool` and `--forbid-tool` unchanged.

Fixes #135

## Testing

- `go test ./cmd/mcpsnoop -run '^TestCheck(MaxDurationAssertion|MaxDurationSummarizesSlowCalls|AssertionsCompose)$' -count=1`
- `make check`
- `go build ./...`
- `git diff --check`

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,7 +331,7 @@ other and with `--fail-on`, and any failure exits non-zero.
 
 | Flag | Fails when |
 |---|---|
-| `--max-duration <dur>` | a completed tool call took longer than the budget, e.g. `--max-duration 500ms` |
+| `--max-duration <dur>` | one or more completed tool calls exceeded the budget; reports their count and the worst call |
 | `--expect-tool <name>` | the named tool was never called (repeatable) |
 | `--forbid-tool <name>` | the named tool was called (repeatable) |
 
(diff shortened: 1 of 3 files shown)


kerlenton:
Looks good to me. Thanks!

PR #143, opened 2026-07-22
feat: gate checks on deprecated protocol features

### LINKED ISSUE #134: Let check gate on deprecated protocol features
## Context

Flagging deprecated Roots, Sampling and Logging landed without a CI signal, deliberately, because whether a build should fail on them is a separate decision. It is worth making now. A team migrating off those features wants a gate that stays red until the traffic is clean, and the data is already captured, the event carries a `Deprecated` field.

The deprecation window is at least twelve months, so this is a migration aid with a known horizon rather than an alarm.

## Proposed change

Add `deprecated` to the `--fail-on` set, opt in, leaving the default `error,invalid,warn` gate untouched. Mechanically it mirrors `drift`, a signal constant, a counter on `checkSummary`, a branch in `count()` which is the single decision point that `failed()` and the JUnit writer both read, and a phrase in `checkSignalFailureReason` so the JUnit case reads properly.

## Acceptance criteria

- [ ] `--fail-on deprecated` fails a session containing a deprecated call
- [ ] A default run is unchanged
- [ ] The signal appears as a JUnit testcase with sensible wording
- [ ] Tests cover selected and not selected

## Files

- `cmd/mcpsnoop/check.go`, `cmd/mcpsnoop/check_junit.go`

AnayGarodia:
## What and why

Adds deprecated protocol features to the existing check gate, selected-text output, summary counts, and JUnit failures. This makes --fail-on deprecated actionable instead of silently accepting deprecated usage.

Closes #134

## Testing

- go test ./cmd/mcpsnoop
- go vet ./...
- staticcheck ./...
- gofmt -s check

The full go test ./... reaches unrelated existing failures in internal/tui: TestStreamRowShowsSupersededStatusInWarnStyle and TestStatusStyleWarnsOnTruncated.

## Checklist

- [ ] make check passes (blocked by the unrelated internal/tui failures above)
- [x] Added or updated tests for the change
- [x] README/docs are unchanged because the existing CLI option is now implemented as documented

### DIFF
--- a/cmd/mcpsnoop/check.go
+++ b/cmd/mcpsnoop/check.go
@@ -15,15 +15,16 @@ import (
 type checkSignal string
 
 const (
-	checkError    checkSignal = "error"
-	checkInvalid  checkSignal = "invalid"
-	checkWarn     checkSignal = "warn"
-	checkMismatch checkSignal = "mismatch"
-	checkPending  checkSignal = "pending"
-	checkDrift    checkSignal = "drift"
+	checkError      checkSignal = "error"
+	checkInvalid    checkSignal = "invalid"
+	checkWarn       checkSignal = "warn"
+	checkMismatch   checkSignal = "mismatch"
+	checkPending    checkSignal = "pending"
+	checkDrift      checkSignal = "drift"
+	checkDeprecated checkSignal = "deprecated"
 )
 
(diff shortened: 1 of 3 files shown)


kerlenton:
Looks good to me. Thanks!

PR #144, opened 2026-07-22
feat(tui): show schema findings in tool summary

### LINKED ISSUE #129: Flag tool schemas that use constructs many clients handle badly
## Context

A server can be perfectly valid and still be hard for an agent to use, and the advice teams keep converging on is to stay inside the plainest part of JSON Schema, flat objects and explicit types, avoiding `$ref` and the composition keywords, on the grounds that a tool which satisfies the strictest validator works everywhere. The 2026-07-28 release moves to full JSON Schema 2020-12, which widens what a server may legally send, and the release notes separately warn implementers not to follow external schema references blindly and to bound validation work.

So the gap widens between what a server may send and what clients reliably accept, and the person debugging a tool the model keeps calling wrongly has no way to see that the schema is the reason.

mcpsnoop already parses and canonicalises every `InputSchema` for drift detection, so the schemas are in hand.

## Proposed change

Flag a tool whose schema uses a construct known to travel badly, and say which construct, in the surface that already lists tools. Worth flagging are the composition keywords, `$ref` in general and an external or remote `$ref` in particular, and a property with no declared type, which is the case that most directly makes a model guess.

An external `$ref` deserves its own wording rather than being folded in with the rest, because it is also the case the spec warns about, a schema that points somewhere off the wire is both a portability problem and something worth noticing.

## Scope boundary

Flag, do not validate. mcpsnoop is not a schema validator and must not resolve a `$ref`, fetch anything, or refuse a tool. It notes that a construct is present and leaves the judgement to the reader. Keep the wording as an observation rather than a rule, since a schema using `oneOf` is not wrong, only likely to be handled inconsistently.

This is about tool definitions only. Do not touch drift detection, which compares schemas and should keep comparing them exactly as it does.

## Acceptance criteria

- [ ] A tool whose schema uses a composition keyword is flagged, naming which
- [ ] A `$ref` is flagged, and an external or remote one is called out separately
- [ ] A property with no declared type is flagged
- [ ] Nothing is resolved or fetched, an external `$ref` is recognised by its form alone
- [ ] A plain schema of flat typed properties is not flagged
- [ ] Drift detection is unchanged
- [ ] Tests cover each construct, an external ref, and a clean schema

## Files

- `internal/store/store.go` or a new file beside it, near the existing schema handling
- `internal/store/views.go`, `ToolDefinition`
- `internal/tui/view.go`, wherever tools are listed

elvinka505:
## What and why

I added schema findings to the tool summary so it is immediately visible when a
tool definition uses JSON Schema constructs that many MCP clients handle
inconsistently.

The analysis is purely observational. It flags composition keywords, internal
and external `$ref`s, and untyped properties without validating, resolving, or
fetching anything. Findings are shown in a dedicated SCHEMA column.

Fixes #129

## Checklist

- [ ] Updated the README / docs if user-facing behaviour changed

### DIFF
--- a/internal/store/schema.go
+++ b/internal/store/schema.go
@@ -5,6 +5,9 @@ import (
 	"strings"
 )
 
+// SchemaFindingKind names a JSON Schema construct that clients handle
+// inconsistently. A finding is an observation, not a verdict: a schema using
+// oneOf is not wrong, only likely to be interpreted differently across clients.
 type SchemaFindingKind string
 
 const (
@@ -21,6 +24,14 @@ type SchemaFinding struct {
 	Kind SchemaFindingKind
 }
 
+// analyzeSchema reports the constructs a tool's advertised input schema uses
+// that are known to travel badly. Nothing is resolved or fetched: an external
+// $ref is recognized by its form alone.
+//
+// Findings are deduplicated by kind. A finding carries only its kind, so two
+// entries of the same kind are indistinguishable and add nothing; collapsing
+// them lets a caller treat "more than one finding" as "more than one kind of
+// problem", which is the question a reader actually has.
 func analyzeSchema(raw json.RawMessage) []SchemaFinding {
 	if len(raw) == 0 {
 		return nil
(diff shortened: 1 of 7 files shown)


kerlenton:
I like it. Thanks!

PR #145, opened 2026-07-22
feat(store): flag invalid MRTR requestState echoes

### LINKED ISSUE #133: Flag a client that does not echo requestState untouched
## Context

MRTR carries server state through the client. When a server answers with an `InputRequiredResult` it may include `requestState`, an opaque string, and the specification is unusually firm about what happens next. The client **MUST** echo the exact value back on the retry, and **MUST NOT** inspect, parse, modify or make any assumption about its contents. If the result carried no `requestState`, the client **MUST NOT** invent one.

The specification also tells servers to treat it as attacker-controlled input, to protect its integrity, and to reject state that fails verification, because a client that tampers with it can try to alter server behaviour or bypass authorization.

That makes the echo a rule that can be checked from traffic alone, and mcpsnoop is the only thing positioned to check it, since it sees the value leave and come back. A client quietly re-encoding or truncating it is a real bug that produces a confusing server-side rejection, and today nothing anywhere would say why.

## Proposed change

Remember the `requestState` of an `InputRequiredResult` and compare it with the value echoed on the retry it is linked to. Flag a mismatch, an absent echo where one was required, and an invented state where none was issued. Say which of the three it is, since the causes differ.

Depends on the MRTR correlation work, since without the link there is nothing to compare against.

## Scope boundary

Compare, do not interpret. mcpsnoop must not parse, decode or validate the contents, that is exactly what the specification forbids the client from doing and there is no reason for an observer to do it either. Comparing opaque bytes is the whole check.

Never render the value itself in full, it may be an encrypted blob carrying a principal and a token. Show that it differs, not what it contains.

## Acceptance criteria

- [ ] A retry echoing a different `requestState` is flagged
- [ ] A retry omitting a `requestState` that was issued is flagged
- [ ] A retry carrying one that was never issued is flagged
- [ ] A correct echo produces nothing
- [ ] The value is never rendered in full anywhere
- [ ] Nothing decodes or parses the contents
- [ ] Tests cover all four cases

## Files

- `internal/store/store.go`, alongside the MRTR correlation
- `internal/tui/view.go` for the marker

jstar0:
## What and why

MRTR requires a client to echo an optional opaque `requestState` unchanged and
to omit it when the server did not issue one. A changed, missing, or invented
state currently prevents exact retry correlation, so mcpsnoop opens the retry
as an unrelated call and cannot explain the resulting server rejection.

This change preserves whether the field was present, keeps exact state matching
as the primary correlation path, and uses the existing method, operation, and
complete response-key evidence only when it identifies one parked call. The
linked retry is classified as changed, missing, or invented; correct echoes stay
clean, and ambiguous traffic remains unlinked. Findings contain only the
category, never the opaque value, and appear through the existing warning path
with a compact `state!` stream marker.

Fixes #133

## Checklist

- [x] README / docs are unchanged; the finding uses the existing per-frame warning and status surfaces

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -341,6 +341,11 @@ func (s *Store) Ingest(e proxy.Envelope) EventView {
 			ev.warning = validationWarning(msg)
 			ev.mrtrStateIssue = stateIssue
 			if warning := stateIssue.warning(); warning != "" {
+				// The spec makes the echo a MUST for the client, so a violation is a
+				// protocol error on the wire rather than an observation of ours. It
+				// therefore rides the warning field like any other protocol warning,
+				// which means a default check run fails on it. That is deliberate: a
+				// client mangling server state is worth stopping a build for.
 				ev.warning = appendWarning(ev.warning, warning)
 			}
 			sess.requests++ // a request on the wire, even though not a new call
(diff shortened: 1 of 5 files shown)


kerlenton:
LGTM! Thanks!

PR #146, opened 2026-07-23
refactor(check): reword the max-duration summary line

### LINKED ISSUE #135: --max-duration prints one line per slow call and floods the CI log

syf2211:
## Summary

Collapse `--max-duration` assertion output so a session with many slow tool calls produces one summary line instead of flooding CI logs with one line per call.

## Motivation

`checkAssertions.eval` previously appended a failure message for every completed tool call over the latency budget. Sessions with dozens or hundreds of slow calls could emit equally many `assertion failed:` lines, burying the signal that matters (how many calls breached the budget and which was worst).

Fixes #135.

## Changes

- Collect all over-budget tool calls in `checkAssertions.eval` and emit a single message naming the count and worst offender.
- Keep existing filtering: only completed, non-superseded tool calls with a real response duration are judged.
- Leave `--expect-tool` and `--forbid-tool` assertions unchanged (still one line each).
- Update `TestCheckMaxDurationAssertion` expectations and add `TestCheckMaxDurationAssertionCollapsesManySlowCalls`.

Example output:
```
assertion failed: 3 tool calls exceeded the 500ms budget (worst: tool "delete" took 3s)
```

## Tests

```bash
GOTOOLCHAIN=go1.26.0 go test -race -run 'TestCheckMaxDuration|TestCheckAssertionsCompose' ./cmd/mcpsnoop/
```

All three tests pass.

`make check` passes for `cmd/mcpsnoop` and all other packages except two pre-existing failures in `internal/tui` on `main` (`TestStreamRowShowsSupersededStatusInWarnStyle`, `TestStatusStyleWarnsOnTruncated`).

## Notes

- Exit code behavior is unchanged: any over-budget call still fails the check.
- JUnit output inherits the collapsed message via the existing `assertionFailures` path — no separate formatting change needed.

### DIFF
--- a/cmd/mcpsnoop/check.go
+++ b/cmd/mcpsnoop/check.go
@@ -15,15 +15,16 @@ import (
 type checkSignal string
 
 const (
-	checkError    checkSignal = "error"
-	checkInvalid  checkSignal = "invalid"
-	checkWarn     checkSignal = "warn"
-	checkMismatch checkSignal = "mismatch"
-	checkPending  checkSignal = "pending"
-	checkDrift    checkSignal = "drift"
+	checkError      checkSignal = "error"
+	checkInvalid    checkSignal = "invalid"
+	checkWarn       checkSignal = "warn"
+	checkMismatch   checkSignal = "mismatch"
+	checkPending    checkSignal = "pending"
+	checkDrift      checkSignal = "drift"
+	checkDeprecated checkSignal = "deprecated"
 )
 
-var checkSignalOrder = []checkSignal{checkError, checkInvalid, checkWarn, checkMismatch, checkPending, checkDrift}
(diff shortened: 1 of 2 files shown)


kerlenton:
LGTM. Thanks!

PR #175, opened 2026-07-25
fix(proxy): flag stdio and SSE frames truncated at observe cap

### LINKED ISSUE #151: A stdio frame over the size cap is reported as stream corruption instead of truncated
**What happened**

`Envelope.Truncated` is only ever set on the HTTP path, by `bodyTap`
(`internal/proxy/http.go`). In `pumpFrames` the observed copy stops growing at
`maxFrameBytes`:

```go
if len(pending) < maxFrameBytes {
    pending = append(pending, chunk...)
}
```

but nothing marks the envelope. The truncated bytes are no longer valid JSON, so
`splitObserved` routes them to `Text`, and the store classifies the frame as
`EventInvalid` — the "server printed a stray line to stdout and corrupted the
stream" diagnosis.

So a valid 16+ MiB `tools/call` result (a large file read, a base64 image) is
reported as stream corruption, and `check --fail-on invalid` fails on it. The
store already has the correct handling for this case at
`internal/store/store.go:273`; the stdio path just never reaches it.

**Expected**

Set `Truncated` when `pumpFrames` caps its copy, same as the HTTP tap.

**Also**

`sseTap` has the same gap: when `dataBuf` hits `maxFrameBytes` it silently drops
the rest and calls `emitFrames(…, route{})` with `truncated: false`.

**Environment**

- v0.14.0, stdio shim

syf2211:
## Summary

Mark stdio and SSE observations as `Truncated` when mcpsnoop caps its copy at the frame-size limit, matching the existing HTTP `bodyTap` behavior.

## Motivation

Oversized stdio JSON-RPC lines and large SSE events were forwarded correctly but their capped observed copies were classified as stream corruption (`EventInvalid`) during `check`, because `Truncated` was never set on those paths.

Fixes #151

## Changes

- `pumpFrames`: track when the observed copy exceeds `frameObserveCap` and emit `Envelope.Truncated` instead of routing through `splitObserved`
- `sseTap`: set `truncated` when the event payload exceeds `sseDataObserveCap` and emit a single flagged frame (same as `observeBody` for HTTP)
- Fix newline stripping when the observed copy was capped before the line terminator was buffered
- Add regression tests for both paths

## Tests

- `go test -race ./internal/proxy/...` — PASS
- `make check` — all packages pass except two pre-existing TUI style tests on `main` (`TestStreamRowShowsSupersededStatusInWarnStyle`, `TestStatusStyleWarnsOnTruncated`)

## Notes

- Observation remains best-effort; the proxied byte stream is unchanged
- Reviewed with composer-2.5 — APPROVE

### DIFF
--- a/internal/proxy/http.go
+++ b/internal/proxy/http.go
@@ -229,8 +229,14 @@ func httpProxyHandler(target *url.URL, emit func(Direction, []byte, route)) http
 			}
 			if strings.HasPrefix(resp.Header.Get("Content-Type"), "text/event-stream") {
 				// Streaming case, tap SSE data frames as bytes flow to the client.
-				resp.Body = newSSETap(resp.Body, func(data []byte) {
-					emitFrames(emit, ServerToClient, data, rt)
+				resp.Body = newSSETap(resp.Body, func(data []byte, truncated bool) {
+					r := rt
+					if truncated {
+						r.truncated = true
+						emit(ServerToClient, data, r)
+						return
+					}
+					emitFrames(emit, ServerToClient, data, r)
 				})
 				return nil
 			}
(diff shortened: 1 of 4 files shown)


kerlenton:
Good work! Thanks!

PR #177, opened 2026-07-25
feat(export): preserve W3C trace context in OTLP spans

### LINKED ISSUE #158: OTLP export ignores W3C Trace Context, so spans never join the caller's trace
**The problem**

`WriteOTLP` synthesises a trace id from the session id:

```go
traceID := otlpID(16, "trace", data.Session.ID)   // internal/exporter/exporter.go:383
```

and `otlpSpan` has no `parentSpanId` field at all. Every mcpsnoop span therefore
lands in a trace of its own.

SEP-414 now documents W3C Trace Context propagation in `_meta` and fixes the key
names `traceparent`, `tracestate` and `baggage` precisely so a trace that starts
in a host application follows the call through the client SDK, the MCP server and
whatever the server calls downstream. mcpsnoop sits in the middle of that path
and is the only component that sees both halves, but it throws the linkage away.

**What you'd like**

When a request's `_meta` carries `traceparent`, parse the trace id and parent
span id out of it and use them for the emitted span instead of the synthetic
ones. Fall back to today's behaviour when absent. Applies to both
`mcpsnoop export -T otlp` and the live `--otlp-endpoint` sink.

That turns the OTLP output from "a standalone picture of one session" into "the
missing MCP layer of a trace you already have", which seems like the whole point
of shipping OTLP.

**Where**

`internal/exporter/exporter.go` (`WriteOTLP`, `otlpSpan`), `internal/otlpsink/sink.go`.

shixi-li:
## What and why

Use W3C Trace Context from each request's `_meta.traceparent` when exporting correlated calls as OTLP. Valid context now supplies the trace ID and parent span ID for both `mcpsnoop export -T otlp` and the live `--otlp-endpoint` sink; missing or invalid context keeps the existing session-derived trace and omits the parent.

The parser validates the fixed W3C fields per call, preserves forward compatibility for later versions, and leaves `tracestate` and `baggage` out of scope.

Closes https://github.com/kerlenton/mcpsnoop/issues/158

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,10 +253,13 @@ mcpsnoop export -T json|html|text|har|otlp [-o file|-] [session-id|log.jsonl|-]
 | `html` | a self-contained browser file with search and collapsible JSON |
 | `text` | a pretty plain-text dump |
 | `har` | one entry per correlated call, openable in browser devtools and anything else that reads HAR |
-| `otlp` | OTLP JSON with a trace per session and a span per correlated call |
+| `otlp` | OTLP JSON with a span per correlated call; W3C trace context joins caller traces, otherwise one trace is used per session |
 
 MCP is not HTTP, so a HAR entry's URL, status code, and timings are a deliberate
 mapping of each call rather than a wire transcript.
+
+For OTLP, a request's `_meta.traceparent` supplies that call's trace and parent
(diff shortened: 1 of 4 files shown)


kerlenton:
I like it! Thanks!

PR #178, opened 2026-07-26
fix(check): report dropped frames in check and export

### LINKED ISSUE #165: `check` and `export` don't report dropped frames, so CI passes on an incomplete capture
**What happened**

The store infers dropped envelopes from gaps in the monotonic per-session `Seq`
and exposes the count as `SessionHeader.MissingFrames`
(`internal/store/views.go:109`). It is rendered in the TUI status bar and
**nowhere else**:

```
$ grep -rn MissingFrames --include='*.go' .
internal/tui/view.go:208
internal/tui/view.go:229
internal/tui/view.go:231
internal/tui/view.go:234
internal/store/views.go:109
internal/store/views.go:246
```

So `mcpsnoop check artifacts/session.jsonl` reports
`errors=0 invalid=0 warnings=0 … check passed` on a trace that is known to be
missing frames — and the shim already knows it happened, since it prints
`dropped N envelope(s) under load` to stderr on shutdown.

A gate that reads green on a capture it knows is incomplete is worse than one
that reads red, because the whole value of `check` is the claim that it looked at
everything.

**Expected**

- Add `MissingFrames` to `SessionSummary` in the JSON export (and the HAR/OTLP
  comment, which currently claims one span per correlated call without noting
  the caveat).
- Report it in `check` output, and add an `incomplete` signal to `--fail-on` so
  a job can refuse to grade a lossy trace. Probably opt-in like `drift`, to avoid
  changing the default gate.

**Where**

`cmd/mcpsnoop/check.go` (`summarizeCheck`, `checkSummary`),
`internal/exporter/exporter.go` (`SessionSummary`).

syf2211:
## Summary

Expose `SessionHeader.MissingFrames` in JSON export `SessionSummary` and surface `missing_frames` in `mcpsnoop check` output. Add an opt-in `incomplete` signal to `--fail-on` so CI can reject lossy captures.

## Motivation

The store already infers dropped envelopes from Seq gaps and shows the count in the TUI, but `check` and `export` omitted it. A gate that reads green on a capture known to be missing frames is worse than one that reads red.

Fixes #165

## Changes

- Add `missing_frames` to exported `SessionSummary` JSON
- Report `missing_frames=N` in `mcpsnoop check` text output
- Add opt-in `incomplete` signal to `--fail-on` (JUnit testcase included)
- Note incomplete-capture caveat in HAR/OTLP export comments
- Update README `--fail-on` documentation

## Tests

- `go test ./cmd/mcpsnoop/... ./internal/exporter/... -count=1` — PASS
- Added tests for default reporting, `--fail-on incomplete`, and JSON export round-trip

## Notes

`incomplete` is opt-in like `drift` and `pending`, so default CI gates are unchanged. Composer-2.5 review: APPROVE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,10 @@ the spec has deprecated, or `incomplete` to gate on captures with dropped frames
 Pass a comma-separated subset to select only the
 conditions relevant to a job. Omit the session to check the newest capture, or use
 `-` to read JSONL from stdin.
+The dropped-frame count travels with the artifacts too, so a capture that
+understates itself says so wherever it is opened: `missing_frames` in the JSON
+export, `log.comment` in HAR, and the `mcpsnoop.session.missing_frames` resource
+attribute in OTLP.
 A name or resource URI that will not fit in an HTTP field value travels Base64 in
 a `=?base64?…?=` sentinel; it is decoded before the comparison, so a client that
 encodes correctly is never flagged.
--- a/README.md
+++ b/README.md
@@ -321,18 +321,20 @@ regression too. Improvements (added tools, fixed calls, speedups) still exit zer
 ## Checking sessions in CI
 
 Gate a recorded agent run on errors, stream corruption, protocol warnings,
-routing-header mismatches, calls that never got a response, tool-definition
-drift, or use of deprecated protocol features.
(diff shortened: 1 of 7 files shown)


kerlenton:
LGTM! Thank you!

PR #179, opened 2026-07-26
fix(store): warn when 2026-07-28 results omit resultType

### LINKED ISSUE #163: Add a conformance check for results missing the required `resultType`
**The problem**

In the draft schema, `Result.resultType` is non-optional with the comment:

> Servers implementing this protocol version **MUST** include this field. For
> backward compatibility, when a client receives a result from a server
> implementing an earlier protocol version (which does not include
> `resultType`), the client **MUST** treat the absent field as `"complete"`.

mcpsnoop already reads `resultType` in `parseInputRequired` and `parseTaskState`,
but only to recognise `input_required` and task handles. A server on 2026-07-28
that omits the field entirely is silently non-conformant, and the client-side
fallback means nothing visibly breaks — which is exactly the sort of quiet drift
this tool is good at catching.

**What you'd like**

When the session's negotiated version is 2026-07-28 or later, warn on a result
frame with no `resultType`. Version-gated, so 2025-11-25 traffic stays clean.

**Where**

`internal/store/store.go`, `validationWarning`.

syf2211:
## Summary

Add a version-gated conformance warning when a server on the 2026-07-28 MCP revision returns a result object without the required `resultType` field.

## Motivation

The 2026-07-28 schema makes `resultType` mandatory on results. mcpsnoop already reads `resultType` for MRTR and task handling, but silently accepted absent fields — the same kind of quiet drift this tool is meant to catch.

Fixes #163

## Changes

- Add `missingResultTypeWarning` and call it from the response ingest path after `validationWarning`
- Gate on session negotiated version or `MCP-Protocol-Version` header (2026-07-28+), matching routing-header checks
- Add `TestMissingResultTypeWarning` with four cases: warn on 2026-07-28, silent on 2025-11-25, clean conforming result, header-only gating

## Tests

- `go test ./internal/store/... -count=1` — PASS
- Composer-2.5 review: APPROVE

## Notes

Warning-only; no change to call tracking or error counts. Discover responses that establish version via `supportedVersions` without prior client _meta are not flagged on that same frame (version is applied in `completeCall` after validation).

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -136,6 +136,15 @@ type call struct {
 	// opName is the tool, prompt or resource this call names, kept so a retry can
 	// be matched back to it without re-parsing the original params.
 	opName string
+	// protocolVersion is the revision this one request declared, which is the only
+	// version a response to it may be judged by. MCP is stateless: the spec says
+	// outright that a server must not rely on prior requests over the same
+	// connection to establish the protocol version, and that clients may interleave
+	// unrelated requests on one transport. Reading the session's version instead
+	// would judge this response by whatever the most recent unrelated request
+	// happened to declare, which flags a correct older answer and excuses a
+	// genuinely non-conformant newer one. Empty when the request declared none.
+	protocolVersion string
 	// mrtrState, its presence bit, and mrtrKeys park what an InputRequiredResult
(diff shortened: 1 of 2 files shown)


kerlenton:
LGTM! Thanks!

PR #181, opened 2026-07-26
fix(store): treat subscriptions/listen as streaming, not pending

### LINKED ISSUE #150: `subscriptions/listen` never completes, so it shows PENDING forever and fails `check --fail-on pending`
**What happened**

The 2026-07-28 revision replaced `resources/subscribe` and the standalone GET
stream with a single long-lived `subscriptions/listen` request. Its response
arrives only when the stream ends, which may be hours later or never.

`openCall` treats it as an ordinary request, so:

- the TUI renders it as a hung call with a live `PENDING` timer, which is exactly
  the signal the hung-call feature exists to make trustworthy;
- `mcpsnoop check --fail-on pending` fails on a perfectly healthy session;
- `session.pending` is permanently off by one per open stream.

**Expected**

Recognise `subscriptions/listen` as a stream-opening request with its own state
rather than a pending call. Two follow-ons that belong with it:

1. Notifications delivered on the stream carry
   `params._meta["io.modelcontextprotocol/subscriptionId"]`, whose value is the
   JSON-RPC id of the `listen` request. Correlating on that would let the stream
   read as one operation, the way MRTR retries and task-backed calls already do.
2. On stdio the *server* sends `notifications/cancelled` referencing the listen
   id solely to terminate the stream. That should close the call, not read as a
   client cancelling its own request.

**Where**

`internal/store/store.go` (`openCall`, `Ingest`), `internal/tui/view.go` status column.

**Environment**

- v0.14.0

syf2211:
## Summary

Recognise `subscriptions/listen` as a long-lived stream request instead of a hung pending RPC. Open listen streams no longer increment the session pending counter, render as `streaming` in the TUI, or fail `mcpsnoop check --fail-on pending` on healthy sessions.

## Motivation

The 2026-07-28 revision replaced resource subscribe streams with a single long-lived `subscriptions/listen` request whose response arrives only when the stream ends. mcpsnoop treated it like any other request, so it showed a live pending timer and permanently inflated `session.pending`.

Fixes #150.

## Changes

- Add a `Streaming` call state for stream-opening requests.
- Open `subscriptions/listen` in `Streaming` state without bumping `pending`.
- Complete streaming calls on the final response; only decrement `pending` when the call was actually pending.
- Close streaming calls when the server sends `notifications/cancelled` for the listen id.
- Show `streaming` status in the TUI and export layer instead of the pending spinner.
- Add store regression tests for pending accounting, normal completion, and cancel notification handling.

## Tests

```bash
make check
```

- `go vet`, `staticcheck`: pass
- `go test -race ./...`: pass (store tests including new subscriptions/listen cases; two pre-existing TUI test failures on `main` unrelated to this change)

## Notes

- Subscription notification correlation via `_meta["io.modelcontextprotocol/subscriptionId"]` is left for a follow-up, as described in the issue.
- OTLP/HTML export could surface `streaming` more explicitly in a later PR.

### DIFF
--- a/internal/exporter/exporter.go
+++ b/internal/exporter/exporter.go
@@ -723,6 +723,8 @@ func exportCall(index int, c store.CallView) CallExport {
 	switch {
 	case c.State == store.Pending:
 		status = "pending"
+	case c.State == store.Streaming:
+		status = "streaming"
 	case c.State == store.Superseded:
 		status = "superseded"
 	case c.TaskStatus == "cancelled":
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -34,6 +34,9 @@ const (
 	// Superseded means the request's id was reused by a later in-flight request, so
 	// this one can never be matched to a response. It is no longer pending.
 	Superseded
+	// Streaming means a long-lived stream request (subscriptions/listen) whose
+	// response arrives only when the stream ends. It is open but not pending.
+	Streaming
 )
 
 func (s CallState) String() string {
@@ -44,6 +47,8 @@ func (s CallState) String() string {
 		return "failed"
 	case Superseded:
 		return "superseded"
+	case Streaming:
+		return "streaming"
 	default:
 		return "pending"
 	}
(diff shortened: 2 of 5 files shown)


kerlenton:
LGTM! Thanks!

PR #183, opened 2026-07-27
feat(http): observe and validate Mcp-Param headers

### LINKED ISSUE #157: `Mcp-Param-{Name}` headers are not observed or validated against the body
**The problem**

2026-07-28 lets a server designate tool parameters to be mirrored into HTTP
headers with an `x-mcp-header` annotation in the tool's `inputSchema`:

```json
"region": { "type": "string", "x-mcp-header": "Region" }
```

producing `Mcp-Param-Region: us-west1` alongside the body. Servers **MUST**
validate these against the body and reject a disagreement with `-32020`, and
intermediaries route and rate-limit on them.

mcpsnoop captures `Mcp-Method`, `Mcp-Name` and `MCP-Protocol-Version` and
nothing else, so this whole surface is invisible.

**What you'd like**

Capture `Mcp-Param-*` on the envelope, and validate each against the value at the
annotated property path in the advertised `inputSchema` — mcpsnoop already holds
those definitions in `session.toolDefinitions`, so it is the one observer that
can do this without asking either side.

Worth handling alongside it: the `=?base64?…?=` sentinel applies to these values
too, and the spec says integers **SHOULD** be compared numerically (`42.0` and
`42` are equal), so a naive string compare would produce false positives.

**Alternatives considered**

Just displaying them without validating would already help, and is a reasonable
first cut.

jstar0:
## What and why

The 2026-07-28 protocol revision lets a tool mark statically reachable input
properties with `x-mcp-header`, mirroring their values into `Mcp-Param-*`
headers for routing and rate limiting. mcpsnoop already observes fixed routing
headers, but these dynamic headers were invisible and a disagreement with the
JSON body could pass through a capture without a finding.

This change captures `Mcp-Param-*` headers on Streamable HTTP requests without
altering forwarded headers or body bytes. When a server advertises a tool,
mcpsnoop precomputes its valid `x-mcp-header` property paths; on `tools/call`, it
compares the observed headers with the body using the protocol's string,
boolean, Base64-sentinel, and safe-integer rules. Recognized missing,
unexpected, malformed, or disagreeing values use the existing warning and
routing-mismatch signal.

Validation stays conservative: it follows only static `properties` paths,
ignores unknown headers, and does not run for stdio, batches, old revisions, or
calls without a matching tool definition. The inspector displays captured
parameter headers, and existing key- and value-based redaction applies before
they reach sinks. The additive envelope field is omitted from older or non-HTTP
frames, so existing session logs continue to decode unchanged.

Fixes #157.

## Verification

```bash
make check
```

The complete formatting, vet, staticcheck, and race-test gate passes.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,9 +350,14 @@ The same signal covers a required routing header that is missing entirely. In
 request with `400` and `-32020`. mcpsnoop raises it only once the session is
 known to speak that revision or later: earlier revisions do not define these
 headers at all, so omitting them there is correct.
-A server's own `-32020` rejection counts as the same signal. The server validates
-`Mcp-Param-{Name}` values and header encodings that mcpsnoop never sees, so its
-verdict is sometimes the only evidence a mismatch happened.
+A server's own `-32020` rejection counts as the same signal. On HTTP
+`tools/call` requests, mcpsnoop also shows each `Mcp-Param-{Name}` header and,
+when the matching advertised tool definition is known, compares it with the
+annotated argument path. Nested properties, the Base64 sentinel, booleans, and
+numeric-equivalent safe integers are handled without string-comparison false
+positives. Unknown parameter headers and sessions without a matching tool
+definition remain observational only. Existing key- and value-based redaction
+also applies to captured parameter-header values before they reach a sink.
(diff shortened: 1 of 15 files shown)


kerlenton:
Great! Thank you!

PR #185, opened 2026-07-29
fix(cli): reject unsafe --label values before session log paths

### LINKED ISSUE #155: `--label` is not sanitised and reaches the session log and baseline paths
**What happened**

`labelFor` strips path separators when it derives a label from the wrapped
command (`strings.LastIndexAny(pick, "/\\")`), but an explicit `--label` is used
verbatim. It flows into:

- `paths.SessionLogPath(sessionID)` → `filepath.Join(SessionsDir(), sessionID+".jsonl")`
- the session id shown and matched everywhere in the TUI and `export`/`open`

so `mcpsnoop --label ../../evil -- node server.js` writes its trace outside the
sessions directory. Same for `mcpsnoop http --label`.

This is the user's own flag, so it isn't a privilege boundary, but a config file
is often generated or shared, and a label with a `/` in it silently produces a
session that `mcpsnoop open <id>` cannot find again.

**Expected**

Reject or sanitise labels containing path separators, `..`, or NUL — with a clear
error, since `--label` is documented as needing to be stable and unique for
baselines.

**Where**

`cmd/mcpsnoop/main.go`, `internal/paths/paths.go`.

syf2211:
## Summary

Reject `--label` values (and derived labels) that contain path separators, traversal segments, or NUL bytes before constructing session ids used as log file names.

## Motivation

Fixes #155. An explicit `--label ../../evil` flowed verbatim into `newSessionID`, so `SessionLogPath` could write traces outside the sessions directory. The same applied to `mcpsnoop http --label`.

## Changes

- Add `paths.ValidateLabel` to reject empty labels, `.` / `..`, `/` / `\`, and NUL bytes
- Validate explicit `--label` in the root command before spawning the shim
- Validate resolved labels in `runShim` after `labelFor` (covers derived labels like `..`)
- Validate labels in the `http` subcommand before session creation
- Add unit and CLI tests for unsafe explicit, derived, and http labels

## Tests

- `go test -race ./cmd/mcpsnoop/... ./internal/paths/...` — pass
- `go vet ./...` and `staticcheck ./...` — pass
- Note: `internal/tui` has two pre-existing failures on `main` (`TestStreamRowShowsSupersededStatusInWarnStyle`, `TestStatusStyleWarnsOnTruncated`), unrelated to this change

## Notes

- Tool baselines hash labels, so they were never vulnerable to path traversal
- `--trace-file` remains a separate, documented override for custom log paths

### DIFF
--- a/cmd/mcpsnoop/main.go
+++ b/cmd/mcpsnoop/main.go
@@ -277,6 +277,11 @@ Repeated shim flags can live in a .mcpsnoop.toml file in the current directory.`
 			if err != nil {
 				return err
 			}
+			if label != "" {
+				if err := paths.ValidateLabel(label); err != nil {
+					return fmt.Errorf("invalid --label: %w", err)
+				}
+			}
 			return codeOf(runShimFn(args, label, traceFile, noTrace, redactConfig(redactSecrets, redactKeys, redactValues, redactPaths), trace))
 		},
 	}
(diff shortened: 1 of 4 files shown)


kerlenton:
Closing this in favour of #187, which is now merged. Thank you for the work, and for the diagnosis, which was right.

PR #186, opened 2026-07-29
fix(store): recognise subscriptions/listen as a stream, not a pending call

### LINKED ISSUE #150: `subscriptions/listen` never completes, so it shows PENDING forever and fails `check --fail-on pending`

AnayGarodia:
## What

The 2026-07-28 revision replaced `resources/subscribe` and the standalone GET stream with a single long-lived `subscriptions/listen` request whose response arrives only when the stream ends, which may be hours later or never. `openCall` treated it as an ordinary request, so:

- the TUI rendered a healthy stream as a hung call with a live `PENDING` timer, which is exactly the signal the hung-call feature exists to keep trustworthy;
- `mcpsnoop check --fail-on pending` failed a perfectly healthy session;
- `session.pending` was permanently off by one per open stream.

## How

- A `subscriptions/listen` request now opens in its own `Listening` call state. It never joins the pending counter, the stream row reads `listening` (cyan, with a live elapsed timer), and `status:listening` finds it while `status:pending` deliberately does not.
- Notifications delivered on the stream carry the listen request's id in `params._meta["io.modelcontextprotocol/subscriptionId"]`; they now correlate back to the listen call through new `SubscriptionID`/`SubscriptionCall` event fields, the way MRTR retries and task-backed calls already read as one operation. The id is compared as raw JSON bytes, since it is the JSON-RPC id of the listen request and that is how call ids are keyed.
- On stdio the server terminates the stream by sending `notifications/cancelled` referencing the listen id. That now settles the listen call as completed (its duration spanning the stream's whole life) rather than reading as a client cancelling its own request. The lookup is gated on the method and the `Listening` state, so an ordinary cancelled call keeps today's meaning.
- The eventual response completes the call without decrementing the pending counter it never incremented, so an unrelated in-flight call stays accounted for.
- Exports agree: the JSON/HAR/HTML status is `listening`, and the HTML filter and request-row status surface it like the TUI.

## Tests

- store: listen opens Listening not pending; stream notifications link to the listen call; server-sent `notifications/cancelled` closes the stream without counting an error; an ordinary call named by `notifications/cancelled` stays pending; the late response completes the stream without underflowing the pending counter.
- check: `--fail-on pending` passes over an open stream and still fails on a genuinely hung call.
- tui: the stream row reads `listening`, matched by `status:listening` and not by `status:pending`.

`gofmt`, `go vet ./...`, and `go test ./...` are clean.

Closes #150

### DIFF
--- a/cmd/mcpsnoop/check_test.go
+++ b/cmd/mcpsnoop/check_test.go
@@ -605,6 +605,25 @@ func TestCheckJUnitReportsSelectedDeprecatedFeature(t *testing.T) {
 	}
 }
 
+// A subscriptions/listen stream stays open for the life of a healthy session
+// (2026-07-28), so it must not trip a pending gate; an ordinary unanswered
+// request still must.
+func TestCheckPendingGateIgnoresOpenListenStream(t *testing.T) {
+	t.Setenv("MCPSNOOP_HOME", t.TempDir())
+	listen := checkEnvelope(1, proxy.ClientToServer, `{"jsonrpc":"2.0","id":1,"method":"subscriptions/listen","params":{}}`)
+
+	code, stdout, stderr := executeCheck(t, []string{"--fail-on", "pending", "-"}, encodeCheckLog(t, listen))
+	if code != 0 || stderr != "" {
+		t.Fatalf("an open stream failed the pending gate: code %d, stderr %q\n%s", code, stderr, stdout)
+	}
+
+	hung := checkEnvelope(2, proxy.ClientToServer, `{"jsonrpc":"2.0","id":2,"method":"tools/list"}`)
+	code, stdout, _ = executeCheck(t, []string{"--fail-on", "pending", "-"}, encodeCheckLog(t, listen, hung))
+	if code != 1 || !strings.Contains(stdout, "check failed: pending") {
+		t.Fatalf("a genuinely hung call must still fail: code %d\n%s", code, stdout)
+	}
+}
+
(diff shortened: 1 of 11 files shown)


kerlenton:
Superseded by #181, which landed the same fix for #150. Thanks for the work, and sorry for the duplicate effort

PR #187, opened 2026-07-29
fix(cli): reject --label values that escape the state dir

### LINKED ISSUE #155: `--label` is not sanitised and reaches the session log and baseline paths

AnayGarodia:
## What

`labelFor` strips path separators when it derives a label from the wrapped command, but an explicit `--label` was used verbatim. The label flows into the session id and from there into `paths.SessionLogPath`'s file name, so `mcpsnoop --label ../../evil -- node server.js` wrote its trace outside the sessions directory, and any label with a separator silently produced a session that `mcpsnoop open <id>` could not find again. Same for `mcpsnoop http --label`, and for a label read from a generated or shared `.mcpsnoop.toml`.

## How

`paths.CheckLabel` rejects labels containing path separators, `..`, or a NUL byte with a clear, actionable error, following the `CheckSocketPath` precedent of validating in the package that owns the paths. Both the root command and `http` validate after `applyConfig`, so a config-file label is held to the same rule as the flag, and the shim never starts under a bad label (exit 2, the usage-error code).

Rejecting rather than sanitising is deliberate: `--label` is documented as needing to be stable and unique for baselines, and a silently rewritten label is neither.

Ordinary labels are untouched: dots, spaces, and unicode still pass, and an empty label still falls back to deriving one from the command or target host.

## Tests

- paths: hostile values (`../../evil`, `a/b`, `a\b`, `..`, `prod..`, NUL) are rejected with an error that says why; realistic labels (`index.js`, `my server`, `café`, `v1.2.3`, empty) pass.
- cmd: the root command exits 2 and never runs the shim for a hostile `--label` or a hostile label from `.mcpsnoop.toml`; `http --label` is rejected the same way; a plain label still wraps the command.

`gofmt`, `go vet ./...`, and `go test ./...` are clean.

Closes #155

### DIFF
--- a/cmd/mcpsnoop/main.go
+++ b/cmd/mcpsnoop/main.go
@@ -198,11 +198,15 @@ func redactConfig(commonSecrets bool, keys redactKeysFlag, values redactValuesFl
 
 func main() { os.Exit(execute(os.Args[1:])) }
 
-// runShimFn and runHubFn are indirected so tests can check how the root command
-// routes the wrapped command without spawning a server or launching the TUI.
+// runShimFn, runHubFn and runHTTPFn are indirected so tests can check how a
+// command routes without spawning a server, launching the TUI, or binding a
+// port. The http seam matters most: without it a test that reaches RunHTTP
+// binds the real --listen address and blocks until the package timeout, which
+// fails the whole package rather than the one test.
 var (
 	runShimFn = runShim
 	runHubFn  = runHub
+	runHTTPFn = proxy.RunHTTP
 )
 
 // exitCode carries a command's process exit code out through cobra's error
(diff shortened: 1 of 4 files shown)


kerlenton:
Great! Thank you!

PR #189, opened 2026-07-30
feat(export): redact captured sessions on read

### LINKED ISSUE #166: No way to redact an already-captured session before sharing it
**The problem**

`--redact-secrets` / `--redact-key` / `--redact-value` / `--redact-path` are all
shim-side: `traceSink` wraps the sink with `NewRedactingSink` at capture time.
Once a session log exists, there is nothing that will scrub it.

That is the wrong moment to have to decide. You find the interesting bug in a
trace you captured without redaction, and now you want to attach it to an issue
or hand it to a colleague — and the only options are to hand over raw
credentials, or reproduce the whole session with the flags on and hope the bug
recurs.

**What you'd like**

Accept the same redaction flags on `export` (and `open`), applying `Redactor` to
each decoded envelope on the way through. `internal/proxy/redact.go` already
operates on `Envelope`, so the machinery exists — it just isn't reachable from
the read path.

A `mcpsnoop redact <session> -o clean.jsonl` producing a scrubbed JSONL that
`open`/`check`/`export` all accept would work too, and composes better.

**Alternatives you've considered**

Always redacting by default, which the README rightly avoids — the raw bytes are
the point of a wire capture.

shixi-li:
## Summary

- add the existing redaction flags to `export` and `open`
- apply redaction while decoding saved envelopes, without rewriting the source JSONL
- reject output paths that refer to the input file before truncation
- document that redaction covers JSON-RPC payloads and text, not envelope metadata

## Testing

- focused post-capture redaction and source-immutability regressions
- `go test -race ./cmd/mcpsnoop ./internal/exporter ./internal/proxy -count=1`
- `make check`
- `go build ./...`
- `GOOS=windows GOARCH=amd64 go build ./...`

Closes https://github.com/kerlenton/mcpsnoop/issues/166

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,12 +284,19 @@ mcpsnoop export session.jsonl --redact-secrets --redact-key project_token -o sha
 mcpsnoop open session.jsonl --redact-path '$.params.arguments.password'
 ```
 
-These flags redact JSON-RPC payloads and stderr or other non-JSON text in the
-exported file or the in-memory TUI view. Envelope metadata such as server labels
-and captured HTTP headers is unchanged. The source JSONL is not rewritten, and
-`export` rejects an output that refers to the source file. Still use a separate
-output path, inspect the result before sharing it, and treat redaction as best
-effort.
+These flags rewrite the exported file or the in-memory TUI view, never the
+source JSONL. `export` refuses an output that names the same file as its input,
+and writes through a temporary file that is renamed into place, so a run that
+fails leaves the previous file whole.
+
+What each flag reaches differs, so check the result rather than assuming. All
+four scrub JSON-RPC payloads, and `--redact-key`, `--redact-path` and
+`--redact-secrets` reach only those. Only `--redact-value` also scrubs stderr,
+other non-JSON text, and the inside of a string. An `Mcp-Param-*` header is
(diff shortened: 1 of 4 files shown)


shixi-li:
Rebased this PR onto current `main` at `2de5b665`; the new head is `d9200a6f`.

Conflict resolution kept both sides' intended behavior:

- retained upstream's `runHTTPFn` test seam and added this PR's `runOpenTUIFn` seam;
- retained upstream's streaming export status while preserving read-time redaction;
- preserved source immutability and the input/output alias guard.

`git range-diff` shows the patch is equivalent apart from that necessary test-seam integration, and the scope remains the same four files.

Validation on the rebased head:

- `make check` — passed (gofmt-s check, `go vet ./...`, `staticcheck ./...`, and `go test -race ./...`);
- focused export/open redaction, source-immutability, same-file/hard-link/stdin alias, default-unredacted, and flag-exposure regressions — passed;
- `go build ./...` — passed;
- `GOOS=windows GOARCH=amd64 go build ./...` — passed;
- `git diff --check` — passed.

The branch is now mergeable. Please review the current head when convenient.

kerlenton:
Looks good. Thank you!

PR #190, opened 2026-07-31
fix(store): parse ttlMs/cacheScope and flag cache misuse

### LINKED ISSUE #160: `ttlMs` and `cacheScope` are not parsed, so cache misuse is invisible
**The problem**

SEP-2549 added `ttlMs` and `cacheScope` to list and resource-read results,
modelled on HTTP `Cache-Control`. mcpsnoop doesn't read either field, so two
concrete, observable problems go unreported — and mcpsnoop is well placed to spot
both, because it sees the timing of every call.

**What you'd like**

Surface the values, and add two observations:

- **A client that ignores `ttlMs`.** Re-issuing `tools/list` well inside the
  freshness window is wasted round trips, and now measurable: mcpsnoop knows when
  the previous response arrived and what TTL it declared.
- **`cacheScope: "public"` on a response that looks user-specific.** The spec is
  explicit that a public response may be cached across authorization contexts by
  any intermediary. A server mislabelling per-user data is a cross-tenant leak,
  and a proxy that sees both the declaration and the payload can flag it.

The first fits the existing observation model (warning colour, no `check` signal).
The second is arguably a real violation and could earn one.

syf2211:
## Summary

Parse SEP-2549 `ttlMs` and `cacheScope` from cacheable MCP list/read responses, surface them on timeline events, and add two observations:

- **Structured observation** when a client re-fetches inside the declared freshness window (`CacheStaleRefetch`, warn colour, does not fail default `check`)
- **Protocol warning** when `cacheScope: "public"` is declared on a `resources/read` response that looks user-specific

## Motivation

mcpsnoop sees the timing of every call and is well placed to spot cache misuse that would otherwise be invisible. Fixes #160.

## Changes

- `internal/store/cache.go` — parse cache hints, track freshness per cache key, detect stale refetch and public-scope mislabelling
- `internal/store/store.go` — session `cacheFresh` state and ingest hooks
- `internal/store/views.go` — `CacheHint` and `CacheStaleRefetch` on `EventView`
- `internal/tui/` — show cache metadata in detail; only stale refetch uses warn styling
- `internal/exporter/exporter.go` — export cache fields
- Tests in `cache_test.go` and `cache_check_test.go`

## Tests

- `go test -race ./internal/store/... ./cmd/mcpsnoop/... ./internal/exporter/...`
- All new cache tests pass; pre-existing TUI colour tests fail in this headless environment on main as well

## Notes

- Public-scope detection is intentionally conservative/heuristic (URI markers + obvious per-user JSON keys)
- Cache hint surfacing is informational in the TUI detail panel; only misuse observations use warn styling

Fixes #160

### DIFF
--- a/cmd/mcpsnoop/cache_check_test.go
+++ b/cmd/mcpsnoop/cache_check_test.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/kerlenton/mcpsnoop/internal/paths"
+	"github.com/kerlenton/mcpsnoop/internal/proxy"
+)
+
+func TestCheckPassesForCacheStaleRefetch(t *testing.T) {
+	t.Setenv("MCPSNOOP_HOME", t.TempDir())
+	now := time.Now()
+	writeCheckLog(t, paths.SessionLogPath("s1"),
+		proxy.Envelope{
+			SessionID: "s1", ServerLabel: "srv", Seq: 1, TS: now, Direction: proxy.ClientToServer,
+			Raw: []byte(`{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28","io.modelcontextprotocol/clientCapabilities":{}}}}`),
+		},
+		proxy.Envelope{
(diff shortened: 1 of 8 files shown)


kerlenton:
Great! Thank you!

PR #209, opened 2026-08-01
fix: validate and separate null request IDs

### LINKED ISSUE #197: A request with "id": null is not flagged, and two of them collide into a bogus id-reuse warning
**The problem**

`"id": null` is not a legal request id in MCP, and mcpsnoop never says so. What it says instead is wrong twice over. It blames the client for reusing an id the client never reused, and it prints one frame's tool name on a different frame's body. The one artifact a developer reaches for when a call vanishes without an answer is the thing pointing at the wrong culprit.

The rule is in the Requests section of the Overview page for 2026-07-28, at /specification/2026-07-28/basic/index.

> * Requests **MUST** include a string or integer ID.
> * Unlike base JSON-RPC, the ID **MUST NOT** be `null`.
> * The request ID **MUST NOT** match the ID of any other request the sender has issued and not yet received a response for.

The first two bullets are word for word identical on /specification/2025-11-25/basic/index, so this check needs no revision gate, unlike the routing header and `resultType` checks.

It happens in the wild. [modelcontextprotocol/python-sdk#2057](https://github.com/modelcontextprotocol/python-sdk/issues/2057), open since 14 February 2026, is titled "Requests with `"id": null` silently misclassified as notifications".

> the request is silently reclassified as a `JSONRPCNotification` and the caller gets a 202 with no response

A caller in that state has a call that never resolves and no error to read, which is exactly what mcpsnoop is for. It is the same family as [modelcontextprotocol/typescript-sdk#2247](https://github.com/modelcontextprotocol/typescript-sdk/issues/2247), "Invalid Request frames silently dropped with no error response — client hangs waiting for id that never resolves", and mcpsnoop already handles that one well, because `validationWarning` flags the missing `jsonrpc` field.

**Reproduction**

Four frames, an `initialize` pair followed by two `tools/call` requests that both carry `"id": null`, one for `read_file` and one for `write_file`.

```jsonl
{"session_id":"n1","server_label":"files","seq":1,"ts":"2026-08-01T10:00:00Z","direction":"c2s","transport":"stdio","raw":{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2026-07-28","_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28","io.modelcontextprotocol/clientCapabilities":{}},"capabilities":{},"clientInfo":{"name":"demo","version":"1.0"}}}}
{"session_id":"n1","server_label":"files","seq":2,"ts":"2026-08-01T10:00:00.5Z","direction":"s2c","transport":"stdio","raw":{"jsonrpc":"2.0","id":1,"result":{"resultType":"complete","protocolVersion":"2026-07-28","capabilities":{"tools":{}},"serverInfo":{"name":"files","version":"0.1"}}}}
{"session_id":"n1","server_label":"files","seq":3,"ts":"2026-08-01T10:00:01Z","direction":"c2s","transport":"stdio","raw":{"jsonrpc":"2.0","id":null,"method":"tools/call","params":{"name":"read_file","arguments":{"path":"/etc/hosts"},"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28","io.modelcontextprotocol/clientCapabilities":{}}}}}
{"session_id":"n1","server_label":"files","seq":4,"ts":"2026-08-01T10:00:02Z","direction":"c2s","transport":"stdio","raw":{"jsonrpc":"2.0","id":null,"method":"tools/call","params":{"name":"write_file","arguments":{"path":"/tmp/out"},"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28","io.modelcontextprotocol/clientCapabilities":{}}}}}
```

On `main` at becf38e.

```
$ mcpsnoop export null-id.jsonl -T text
mcpsnoop session n1 (files)
frames: 4  calls: 3  requests: 3  responses: 1  errors: 0  pending: 1

...

#3 2026-08-01T10:00:01Z c2s request tools/call id=null status=pending duration_ms=pending tool=write_file
{
  "jsonrpc": "2.0",
  "id": null,
  "method": "tools/call",
  "params": {
    "name": "read_file",
    "arguments": {
      "path": "/etc/hosts"
    },
    ...
  }
}

#4 2026-08-01T10:00:02Z c2s request tools/call id=null warning=request reuses an id already in flight status=pending duration_ms=pending tool=write_file
{
  "jsonrpc": "2.0",
  "id": null,
  "method": "tools/call",
  "params": {
    "name": "write_file",
    "arguments": {
      "path": "/tmp/out"
    },
    ...
  }
}
```

Frame #3's body says `read_file` while mcpsnoop's summary line for that same frame says `tool=write_file`.

The run does fail CI, and for the wrong reason.

```
$ mcpsnoop check null-id.jsonl
session n1: errors=0 invalid=0 warnings=1 mismatches=0 pending=1 deprecated=0 missing_frames=0
check failed: warn
exit=1
```

Three separate things are going wrong.

*No warning for the illegal id.* `validationWarning` at internal/store/store.go:1584-1601 checks `jsonrpc` and the result/error pairing, and nothing else. Its only id-aware branch is guarded by `msg.Method == ""` at internal/store/store.go:1592, so a request's id is never inspected at all, whatever it holds.

*A reuse warning for a reuse that did not happen.* `RPCMessage.ID` is a `json.RawMessage` (internal/proxy/frame.go:121), so a null id decodes to the four bytes `null` rather than to nothing, which is confirmable in isolation.

```
{"jsonrpc":"2.0","id":null,"method":"tools/call"} -> ID="null" len=4
{"jsonrpc":"2.0","method":"x"}                    -> ID=""     len=0
```

`len(msg.ID) > 0` therefore holds at internal/store/store.go:487 and the frame is classified as a request, which is right, and `IsNotification` at internal/proxy/frame.go:133 agrees. But `ev.id = string(msg.ID)` at internal/store/store.go:490 makes the correlation id the literal string `null`, and `openCall` at internal/store/store.go:783 keys on `callKey{dir, id, conn}` (internal/store/store.go:113-123). Every null-id request in one direction on one connection lands on one key. The second displaces the first, the first is marked `Superseded`, and the second collects "request reuses an id already in flight" from internal/store/store.go:529. mcpsnoop reports the one rule that was not broken and stays silent about the one that was.

The accounting goes with it. Both calls hang forever, since neither can ever be answered, yet only one is counted.

```
$ mcpsnoop export null-id.jsonl -T json   # calls, abridged
{"id": "1",    "method": "initialize", "state": "completed", "duration_ms": 500}
{"id": "null", "method": "tools/call", "tool_name": "read_file",  "state": "superseded"}
{"id": "null", "method": "tools/call", "tool_name": "write_file", "state": "pending"}
```

`--fail-on pending` sees one, not two.

*The wrong tool name on the frame.* The exporter keeps a call key of its own, `dir + "\x00" + c.ID` at internal/exporter/exporter.go:939-941, and the index map built at internal/exporter/exporter.go:305-310 keeps only the last call written under a key. `exportEvent` resolves every event through that map at internal/exporter/exporter.go:819-823, so both frames come out with the same `call_index`.

```
{"seq": 3, "direction": "c2s", "kind": "request", "method": "tools/call", "id": "null", "call_index": 2}
{"seq": 4, "direction": "c2s", "kind": "request", "method": "tools/call", "id": "null", "warning": "request reuses an id already in flight", "call_index": 2}
```

The text renderer then reads the tool name off `data.Calls[*ev.CallIndex]` at internal/exporter/exporter.go:860-864, which is where `tool=write_file` gets stamped onto frame #3. The HTML export resolves the same way, at internal/exporter/html.go:145 and internal/exporter/html.go:181. The TUI is unaffected, because it reads `e.Call.ToolName` directly at internal/tui/view.go:646-647, and so is the HAR export, which walks the call list rather than the index.

**What you'd like**

1. `validationWarning` reports a request whose id is JSON `null`, naming the rule it breaks, with wording along the lines of `request id is null, and 2026-07-28 requires a string or integer id`. This is decidable from the wire with nothing guessed. The id is four literal bytes on the frame, and there is no question of intent, so it is safe as a warning that fails a default `check` run.

2. Null-id requests stop sharing a correlation key. A null id identifies nothing, so it can correlate nothing. Each such request should get its own call entry, for example keyed on `Envelope.Seq` (internal/proxy/frame.go:62) instead of on the literal id bytes. That removes the false reuse warning, keeps the first call out of `Superseded`, leaves the pending count honest at two, and stops one frame inheriting another frame's method and tool name.

3. The exporter's key needs the same identity. Fixing only the store still leaves both calls colliding at internal/exporter/exporter.go:939-941, so whatever identity the store gives a null-id call has to be the thing the exporter keys on.

4. The same spec sentence rules out a float, boolean, array or object id, and all of those are silent today too. Covering them in the same check is cheap, with wording that names what actually arrived.

```
#1 c2s request ping id=1.5 status=pending duration_ms=pending
#2 c2s request ping id={"a":1} status=pending duration_ms=pending
#3 c2s request ping id=true status=pending duration_ms=pending
```

**Out of scope**

A null id on an *error response* stays legal and must keep working. The same Overview page says of error responses that they "**MUST** include the same ID as the request they correspond to (except in error cases where the ID could not be read due a malformed request)", which is the JSON-RPC 2.0 case for a parse error. Only a frame carrying a `method` should be judged by the new rule. Today that case behaves, and it should stay that way.

```
#1 c2s request tools/list id=1 status=pending duration_ms=pending
#2 s2c response id=null warning=response id has no matching request
```

No reclassification. A null-id frame stays a request on the timeline, since the sender wrote a method and plainly wanted an answer. The job here is to report the violation, not to guess whether a notification was meant. That also keeps `IsNotification` at internal/proxy/frame.go:133 as it is.

No shim change. All of this is hub-side. The shim keeps forwarding bytes without interpreting them.

**Where**

- internal/store/store.go:1584-1601, `validationWarning`, where the new check belongs
- internal/store/store.go:1592, the `msg.Method == ""` guard that keeps request ids out of validation entirely
- internal/store/store.go:487-490, request classification and `ev.id = string(msg.ID)`
- internal/store/store.go:113-123 and internal/store/store.go:783, `callKey` and `openCall`, the shared key
- internal/store/store.go:529, the "request reuses an id already in flight" warning
- internal/proxy/frame.go:121, `RPCMessage.ID` as `json.RawMessage`, why a null id is four bytes and not nothing
- internal/proxy/frame.go:62, `Envelope.Seq`, an identity already on every frame
- internal/exporter/exporter.go:305-310, :819-823, :860-864 and :939-941, the second collision and the mislabelled `tool=`
- internal/exporter/html.go:145 and :181, the same resolution in the HTML export
- internal/tui/view.go:646-647 and the HAR path, unaffected, useful as the contrast when writing the test

Guflly:
## What and why

Warn when requests use IDs outside MCP's string and integer types. Null IDs now remain separate pending calls, and exported events keep the right call and tool name. Closes #197.

## Checklist

- [ ] `make check` passes (Windows vet, staticcheck, build, and the changed package tests pass; the full race suite runs in CI)
- [x] README/docs are not needed for this protocol-validation fix

### DIFF
--- a/internal/exporter/exporter.go
+++ b/internal/exporter/exporter.go
@@ -937,7 +937,7 @@ func eventKind(k store.EventKind) string {
 }
 
 func callKey(c store.CallView) string {
-	return string(c.ReqDir) + "\x00" + c.ID
+	return strconv.FormatUint(c.RequestSeq, 10)
 }
 
 func formatDuration(ms *float64) string {
--- a/internal/exporter/exporter_test.go
+++ b/internal/exporter/exporter_test.go
@@ -163,6 +163,41 @@ func TestBuildExportsSupersededCallNotAsOk(t *testing.T) {
 	}
 }
 
+func TestBuildKeepsNullIDCallsDistinct(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "null-id.jsonl")
+	t0 := time.Date(2026, 8, 1, 10, 0, 0, 0, time.UTC)
+	writeEnv(t, path, proxy.Envelope{
+		SessionID: "s1", ServerLabel: "files", Seq: 1, TS: t0,
+		Direction: proxy.ClientToServer,
+		Raw:       json.RawMessage(`{"jsonrpc":"2.0","id":null,"method":"tools/call","params":{"name":"read_file"}}`),
+	})
+	writeEnv(t, path, proxy.Envelope{
+		SessionID: "s1", ServerLabel: "files", Seq: 2, TS: t0.Add(time.Second),
+		Direction: proxy.ClientToServer,
+		Raw:       json.RawMessage(`{"jsonrpc":"2.0","id":null,"method":"tools/call","params":{"name":"write_file"}}`),
+	})
+
+	st, id, err := LoadFile(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	out, err := Build(st, id)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(out.Calls) != 2 || len(out.Events) != 2 {
+		t.Fatalf("calls/events = %d/%d, want 2/2", len(out.Calls), len(out.Events))
+	}
+	for i, want := range []string{"read_file", "write_file"} {
+		if out.Events[i].CallIndex == nil || *out.Events[i].CallIndex != i {
+			t.Fatalf("event %d call index = %v, want %d", i, out.Events[i].CallIndex, i)
+		}
+		if out.Calls[i].ToolName != want {
+			t.Fatalf("call %d tool = %q, want %q", i, out.Calls[i].ToolName, want)
+		}
+	}
+}
+
 func TestBuildExportsCancelledTaskWithoutFlaggingAnError(t *testing.T) {
 	path := filepath.Join(t.TempDir(), "cancel.jsonl")
 	t0 := time.Date(2026, 7, 6, 12, 0, 0, 0, time.UTC)
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -112,6 +112,7 @@ func httpFailed(status int) bool { return status >= 400 }
 // the opposite direction with the same id.
 type callKey struct {
 	dir proxy.Direction
+	seq uint64
 	id  string
 	// conn separates the id spaces of two clients that share one proxy process.
 	// JSON-RPC scopes id uniqueness to the sender, "the ID MUST NOT match the ID
@@ -124,18 +125,19 @@ type callKey struct {
 
 // call is the mutable internal record for one request/response pair.
 type call struct {
-	id       string
-	method   string
-	reqDir   proxy.Direction
-	params   json.RawMessage
-	result   json.RawMessage
-	err      *proxy.RPCError
-	start    time.Time
-	end      time.Time
-	state    CallState
-	isTool   bool
-	toolName string
-	toolErr  bool // result.isError == true (MCP tool-level failure)
+	requestSeq uint64
+	id         string
+	method     string
+	reqDir     proxy.Direction
+	params     json.RawMessage
+	result     json.RawMessage
+	err        *proxy.RPCError
+	start      time.Time
+	end        time.Time
+	state      CallState
+	isTool     bool
+	toolName   string
+	toolErr    bool // result.isError == true (MCP tool-level failure)
 	// errored is the "something went wrong" axis: a protocol error, a tool-level
 	// error, or a task that ended failed. It drives the session error counter and
 	// the CI gate, and is deliberately distinct from the Failed state, which also
@@ -494,7 +496,7 @@ func (s *Store) Ingest(e proxy.Envelope) EventView {
 			// A continuation, not a new call. Mapping the retry id onto the same
 			// call object lets completeCall find it, so the operation keeps one
 			// pending slot and one duration however many round trips it takes.
-			sess.calls[callKey{dir: e.Direction, id: ev.id, conn: e.ConnID}] = root
+			sess.calls[requestCallKey(ev.id, msg.ID, e)] = root
 			root.mrtrState, root.mrtrKeys = "", ""
 			root.mrtrHasState = false
 			sess.unpark(root)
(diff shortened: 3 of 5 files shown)


kerlenton:
Looks good! Thanks!

PR #211, opened 2026-08-02
fix(store): settle a cancelled call instead of reading it as a clean success

### LINKED ISSUE #198: A cancelled call reads as a clean success, and a conforming cancellation reads as a hung request
**The problem**

When a host gives up on a tool call it sends `notifications/cancelled` and shows the user something like `MCP error -32001: Request timed out`, which points the finger at the server or the network. Very often the server was healthy and answered a moment later, and the host threw the answer away. mcpsnoop is the one thing in the path that saw both frames, and today it reports that call as a clean success and says nothing anywhere about the cancel.

The mirror case fails in the opposite direction. A server that obeys the cancel and never answers leaves the call `Pending` forever, so a conforming cancellation is indistinguishable from a hang.

`notifications/cancelled` is handled at `internal/store/store.go:559-567`, and the only thing it does to the referenced call is `sess.closeStreamingCall`. That function returns without touching anything unless the call is in `Streaming` state (`internal/store/store.go:1454-1464`), and `Streaming` is set only for `subscriptions/listen` (`internal/store/store.go:809-813`). `CallState` at `internal/store/store.go:24-40` has `Pending`, `Completed`, `Failed`, `Superseded` and `Streaming`, and no cancelled state at all. So an ordinary cancelled `tools/call` is never touched by the notification that cancelled it.

Both halves reproduced on a build of `main` at `becf38e`, with a stdio server that answers `slow_job` after 600ms and a client that cancels 160ms in.

```python
# server.py, the part that matters
elif method == "tools/call":
    threading.Thread(target=slow, args=(msg["id"], 0.6), daemon=True).start()
elif method == "notifications/cancelled":
    cancelled.add(msg.get("params", {}).get("requestId"))   # obeyed only when OBEY_CANCEL=1
```

```python
# client.py, the part that matters
send({"jsonrpc": "2.0", "id": 3, "method": "tools/call", "params": {"name": "slow_job", "arguments": {}}})
time.sleep(0.15)
send({"jsonrpc": "2.0", "method": "notifications/cancelled",
      "params": {"requestId": 3, "reason": "Request timed out"}})
```

Case one, the server ignores the cancel and answers 450ms after it. `mcpsnoop export --format text`

```
#7 2026-08-01T19:29:05.553472+03:00 c2s request tools/call id=3 status=ok duration_ms=610.543 tool=slow_job
{
  "jsonrpc": "2.0",
  "id": 3,
  "method": "tools/call",
  "params": {
    "name": "slow_job",
    "arguments": {}
  }
}

#8 2026-08-01T19:29:05.71364+03:00 c2s notification notifications/cancelled
{
  "jsonrpc": "2.0",
  "method": "notifications/cancelled",
  "params": {
    "requestId": 3,
    "reason": "Request timed out"
  }
}

#9 2026-08-01T19:29:06.164015+03:00 s2c response id=3 status=ok duration_ms=610.543 tool=slow_job
```

`status=ok` on both the request and the response. The cancel is logged as an unrelated notification with nothing linking it to id 3, and the `reason` string the host wrote is not carried onto the call. `mcpsnoop check` on the same log

```
session server.py-98627-69471b1fd107: errors=0 invalid=0 warnings=0 mismatches=0 pending=0 deprecated=0 missing_frames=0
check passed
```

`mcpsnoop export --format json` agrees, so anything reading the export downstream inherits the same wrong story.

```json
{
  "index": 2,
  "id": "3",
  "method": "tools/call",
  "direction": "c2s",
  "state": "completed",
  "status": "ok",
  "is_tool": true,
  "tool_name": "slow_job",
  "is_error": false,
  "tool_error": false,
  "duration_ms": 610.543
}
```

The summary block credits the whole 610ms to a call the host abandoned at 160ms, with nothing saying so.

```json
"tools": [
 {"name": "slow_job", "calls": 1, "errors": 0, "pending": 0,
  "p50_ms": 610.543, "p95_ms": 610.543, "p99_ms": 610.543}
],
"slowest_calls": [
 {"call_index": 2, "id": "3", "tool_name": "slow_job", "duration_ms": 610.543, "is_error": false}
]
```

Case two, the same traffic with the server obeying the cancel and never answering.

```
#7 2026-08-01T19:29:26.817146+03:00 c2s request tools/call id=3 status=pending duration_ms=pending tool=slow_job
#8 2026-08-01T19:29:26.977361+03:00 c2s notification notifications/cancelled
```

```
$ mcpsnoop check --fail-on error,invalid,warn,pending obey.jsonl
session server.py-98677-a6c1a5d868e3: errors=0 invalid=0 warnings=0 mismatches=0 pending=1 deprecated=0 missing_frames=0
check failed: pending
exit=1
```

`pending` is not in the default `--fail-on` set, which is `error,invalid,warn` (`cmd/mcpsnoop/check.go:141`), so this does not break a default run. It does mean the one signal a user would reach for to catch a hung server cannot be switched on, because the best-behaved server in the world trips it every time a host times out. The TUI shows the same call under a spinner with a live elapsed clock (`internal/tui/view.go:651-654`), which is the "possibly hung" affordance, on a call nobody is waiting for.

The spec, [Cancellation](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/cancellation) for revision 2026-07-28, describes exactly these two shapes and neither is an error.

> 3. Servers receiving cancellation notifications **SHOULD**:
>    * Stop processing the cancelled request
>    * Free associated resources
>    * Not send a response for the cancelled request

> 5. The client **SHOULD** ignore any response to the cancelled request that arrives afterward

and, in Timing Considerations,

> Due to network latency, cancellation notifications may arrive after request processing has completed, and potentially after a response has already been sent.
>
> Both parties **MUST** handle these race conditions gracefully

So the server in case one may have already written the response before it read the cancel, and mcpsnoop cannot see inside the server to know. What mcpsnoop can decide from the wire is the ordering of the two frames at its own vantage point and the gap between them, which is all a reader needs. The spec also asks for exactly this to be recorded, under Implementation Notes.

> Both parties **SHOULD** log cancellation reasons for debugging

This is a live complaint, not a hypothetical. [anthropics/claude-code#58687](https://github.com/anthropics/claude-code/issues/58687) was filed after the reporter built "a stdio tee shim wrapping mcp-remote" by hand to get wire-level evidence, which is mcpsnoop's whole job. Their ask, verbatim, is "When the host does time out and emits `notifications/cancelled`, any `result` that subsequently arrives for the cancelled `id` over the transport should be surfaced as a warning in debug logs at minimum, today it is silently discarded, masking the fact that the work actually completed." A follow-up comment on the same issue calls it "Silent data loss: the server's valid result arrives intact AFTER the host has issued notifications/cancelled", and adds that "Users see '-32001 Request timed out' pointing falsely at the network/server". A second reporter there says "Cowork's MCP client cancels long-running tool calls at ~60s" and "the finished result arrives right after cancellation and is discarded". That issue is closed and is about host timeout policy, which is not mcpsnoop's problem to fix. Making the evidence visible is.

**What you'd like**

Add a `Cancelled` state to `CallState` and settle the referenced call into it, so a cancel stops being invisible on both sides.

On a `notifications/cancelled` naming a request mcpsnoop has open in this session, and only when `closeStreamingCall` did not already claim it, move that call from `Pending` to `Cancelled`, record the cancel timestamp, and release its pending slot the way `Superseded` does at `internal/store/store.go:790-800`. An id mcpsnoop never observed stays a no-op, matching the discipline already written down for `serverCancelWarning` at `internal/store/conformance.go:90-110`, so nothing is invented on evidence mcpsnoop does not have. A `Cancelled` call with no response has no latency, so it should omit the duration exactly as a superseded one does at `internal/exporter/exporter.go:788-796`, and it should not carry the pending spinner in the TUI.

When a response later arrives for a call already `Cancelled`, keep it. The bytes are the evidence and the whole point is that the work completed. Attach a note to the response event carrying the ordering and the gap, something like `result arrived 450ms after the client cancelled this call, and 2026-07-28 says the client should ignore it`. Carry the cancel `reason` string onto the call too, since the spec asks for it to be logged and it is the host's own words about what went wrong.

Make it an observation rather than a warning. A warning fails a default `check` run, and Timing Considerations explicitly blesses the race, so accusing a server of a violation mcpsnoop cannot prove would be wrong. Instead expose the fact structurally so a reader and a CI job can key on it. Concretely, the two things worth landing are a distinct status in `export` for both shapes, cancelled with no result and cancelled with a late result, and a count in the `check` summary line next to `pending`, plus an opt-in `--fail-on` signal for the late-result case for anyone who wants CI to catch it. The default run should stay green on both captures above.

One naming trap. `cancelled` is already taken as an export status and a TUI row status by the Tasks feature, driven by `TaskStatus` from `notifications/tasks` (`internal/exporter/exporter.go:754-758` and `internal/tui/view.go:768-773`). That is a different thing from a wire cancel of a request id, and the two must not collapse into one label.

**Out of scope**

Streamable HTTP. The spec is explicit that "Closing the SSE response stream is the cancellation signal" there and that "No `notifications/cancelled` message is required or expected", so there is no frame to key on and this fix is stdio-shaped by construction. That is a separate problem and it should not be bolted on here.

Also out of scope, changing `--max-duration`. The 610ms in case one is a real server latency and the budget is right to judge it. The gap is that the reader cannot tell the host had already walked away, which the new fields fix without touching the assertion at `cmd/mcpsnoop/check.go:158-200`.

And out of scope, guessing why the host cancelled or whether the server was at fault. mcpsnoop reports the ordering and the delta it observed and stops there.

**Where**

- `internal/store/store.go:24-53`, `CallState` and its `String`, where `Cancelled` goes.
- `internal/store/store.go:559-567`, the `notifications/cancelled` branch that today only calls `closeStreamingCall`.
- `internal/store/store.go:1454-1464`, `closeStreamingCall`, which returns unless the call is `Streaming`.
- `internal/store/store.go:864-878`, `completeCall`, which needs to handle a response landing on a call that is already `Cancelled` rather than treating it as an ordinary pending completion, including the `sess.pending--` accounting.
- `internal/exporter/exporter.go:747-796`, `exportCall`, for the status mapping and the duration omission.
- `internal/tui/view.go:645-663` and `internal/tui/view.go:904-917`, the row status and the row color, so a cancelled call stops rendering as pending or as a plain success.
- `cmd/mcpsnoop/check.go:17-47` and `cmd/mcpsnoop/check.go:105-106`, the signal set and the summary line.

hsdfat:
## What and why

Closes #198.

A cancelled call read as a clean success. `notifications/cancelled` only closed a streaming call, and a late response could still overwrite the outcome as completed or failed.

Adds a `Cancelled` call state, with a reason and timestamp, that:

- settles the call the notification names, unless `closeStreamingCall` already claimed it
- releases the pending slot the way `Superseded` does
- survives a late response instead of being overwritten by it
- shows up in exports, the TUI row status, and the `check` summary

Uses `cancelled_request` / `cancelled_late_result`, since plain `cancelled` is already taken by the Tasks feature. A default `check` run stays green — the late-result case is opt-in via `--fail-on`. An id that was never observed stays a no-op.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,13 +360,14 @@ leave the capture incomplete, tool-definition drift, or use of deprecated
 protocol features.
 
 ```bash
-mcpsnoop check [--format text|junit] [--fail-on error,invalid,warn,mismatch,pending,drift,deprecated,incomplete] [session-id|log.jsonl|-]
+mcpsnoop check [--format text|junit] [--fail-on error,invalid,warn,mismatch,pending,late,drift,deprecated,incomplete] [session-id|log.jsonl|-]
 ```
 
 The three default signals (error, invalid, warn) fail the check. Add `pending`
 to gate on calls that never got a response, `mismatch` to gate specifically on a
-routing header (Mcp-Method or Mcp-Name) disagreeing with the body, `drift` to gate
-on tool definitions changing after approval, `deprecated` to gate on features
+routing header (Mcp-Method or Mcp-Name) disagreeing with the body, `late` to gate
+on a result that arrived after the host had already cancelled the call, `drift` to
+gate on tool definitions changing after approval, `deprecated` to gate on features
 the spec has deprecated, or `incomplete` to gate on captures with dropped frames.
 Pass a comma-separated subset to select only the
 conditions relevant to a job. Omit the session to check the newest capture, or use
@@ -391,6 +392,19 @@ numeric-equivalent safe integers are handled without string-comparison false
 positives. Unknown parameter headers and sessions without a matching tool
 definition remain observational only. Existing key- and value-based redaction
 also applies to captured parameter-header values before they reach a sink.
+
+A `notifications/cancelled` naming an open request settles that call: it stops
+counting toward `pending` (so a conforming cancellation no longer looks like a
+hang), it is counted as `cancelled` on the summary line, and it carries the
+host's `reason` into the export and the TUI. That is an observation, not a
+failure — the specification's Timing Considerations bless the race — so the
+default run stays green whether the server obeyed the cancel or answered anyway.
+The export gives the two shapes their own statuses, `cancelled_request` for a
+call nobody ever answered and `cancelled_late_result` for one the server answered
+after the host had stopped listening, and the response frame carries a
+`late_result` note with the gap. `--fail-on late` is the opt-in gate for that
+second shape, for anyone who wants CI to catch work that completed into a void.
+
 Use `--format junit` to write one JUnit `<testcase>` per signal and session;
 failures follow the same `--fail-on` selection as the text output.
 
--- a/cmd/mcpsnoop/check.go
+++ b/cmd/mcpsnoop/check.go
@@ -15,17 +15,21 @@ import (
 type checkSignal string
 
 const (
-	checkError      checkSignal = "error"
-	checkInvalid    checkSignal = "invalid"
-	checkWarn       checkSignal = "warn"
-	checkMismatch   checkSignal = "mismatch"
-	checkPending    checkSignal = "pending"
+	checkError    checkSignal = "error"
+	checkInvalid  checkSignal = "invalid"
+	checkWarn     checkSignal = "warn"
+	checkMismatch checkSignal = "mismatch"
+	checkPending  checkSignal = "pending"
+	// checkLate is a result that arrived after its call had already been
+	// cancelled. Opt-in only: the spec's Timing Considerations bless the race, so
+	// a default run reports the cancel and stays green.
+	checkLate       checkSignal = "late"
 	checkDrift      checkSignal = "drift"
 	checkDeprecated checkSignal = "deprecated"
 	checkIncomplete checkSignal = "incomplete"
 )
 
-var checkSignalOrder = []checkSignal{checkError, checkInvalid, checkWarn, checkMismatch, checkPending, checkDrift, checkDeprecated, checkIncomplete}
+var checkSignalOrder = []checkSignal{checkError, checkInvalid, checkWarn, checkMismatch, checkPending, checkLate, checkDrift, checkDeprecated, checkIncomplete}
 
 type checkOutputFormat string
 
@@ -41,6 +45,8 @@ type checkSummary struct {
 	warnings        int
 	mismatches      int
 	pending         int
+	cancelled       int
+	lateResults     int
 	deprecated      int
 	missingFrames   uint64
 	drift           store.ToolDrift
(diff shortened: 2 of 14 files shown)


kerlenton:
Closed in #210

PR #212, opened 2026-08-02
feat(check): emit SARIF 2.1.0 so findings land in the Security tab

### LINKED ISSUE #205: Emit SARIF from check so findings land in the Security tab
## Context

`check` is the one command whose job is to turn a build red. It already computes eight signals and three assertions. What it cannot do is put a finding anywhere a person actually reads findings, which on GitHub means the Security tab and the inline pull request annotation. The format that gets content there is SARIF 2.1.0.

Today `--format` accepts two values. `cmd/mcpsnoop/check.go:33-34` declares them, `cmd/mcpsnoop/check.go:142` documents the flag, and `cmd/mcpsnoop/check.go:238` rejects everything else. `cmd/mcpsnoop/check_junit.go` is the only non-text writer in the tree.

Built from `main` at `becf38e` and run against a small stdio server driven so the client reuses an id already in flight, sends a frame carrying neither `result` nor `error`, and the server answers one id twice.

```
$ mcpsnoop check --fail-on error,invalid,warn,pending
session server.py-97119-7cb1b22def44: errors=1 invalid=0 warnings=3 mismatches=0 pending=0 deprecated=0 missing_frames=0
recorded first-seen tool baseline (trusted, not verified)
check failed: error,warn
$ echo $?
1
```

```
$ mcpsnoop check --format sarif
mcpsnoop check: --format must be text or junit, got "sarif"
$ echo $?
2
```

The junit report carries the same aggregate and nothing more. Three distinct protocol warnings collapse into the string `has 3 warnings`.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites name="mcpsnoop check" tests="9" failures="2" errors="0" skipped="0" time="0">
  <testsuite name="server.py-97119-7cb1b22def44" tests="9" failures="2" errors="0" skipped="0" time="0">
    <testcase classname="mcpsnoop.check" name="server.py-97119-7cb1b22def44/error" time="0">
      <failure message="session server.py-97119-7cb1b22def44 has 1 error" type="mcpsnoop.check.error">session server.py-97119-7cb1b22def44 has 1 error</failure>
    </testcase>
    <testcase classname="mcpsnoop.check" name="server.py-97119-7cb1b22def44/invalid" time="0"></testcase>
    <testcase classname="mcpsnoop.check" name="server.py-97119-7cb1b22def44/warn" time="0">
      <failure message="session server.py-97119-7cb1b22def44 has 3 warnings" type="mcpsnoop.check.warn">session server.py-97119-7cb1b22def44 has 3 warnings</failure>
    </testcase>
```

The store already knows which frame caused each of those three. `export --format json` over the same capture prints them one per frame.

```json
[
  {
    "seq": 8,
    "direction": "c2s",
    "warning": "request reuses an id already in flight"
  },
  {
    "seq": 9,
    "direction": "c2s",
    "warning": "response has neither result nor error"
  },
  {
    "seq": 15,
    "direction": "s2c",
    "warning": "duplicate response for the same id"
  }
]
```

So the per-finding data exists and every CI-facing consumer of `check` throws it away. The README's own recipe then ends by handing the report to `actions/upload-artifact` (`README.md:417-427`), which is the honest end state today. The report becomes a zip nobody downloads.

There is prior art in this exact space. MCP Shark advertises "4 output formats: Terminal, JSON, SARIF v2.1.0, HTML" (https://github.com/mcp-shark/mcp-shark). SARIF is what GitHub ingests from any third-party tool, and from the SARIF support page, "Code scanning only supports SARIF version `2.1.0`" (https://docs.github.com/en/code-security/reference/code-scanning/sarif-files/sarif-support).

This stays inside the house constraints. It is a serializer over data `summarizeCheck` and `checkAssertions.eval` already return, it adds no analysis, no heuristic about intent, no state, and no network call of mcpsnoop's own. The upload remains a step in the user's own workflow.

## Decision needed first, what a result points at

A SARIF result is only worth more than a junit failure if it carries a location, and the location has to survive GitHub's path handling. Two things make the obvious answer wrong, so pick a rule before writing the writer.

First, the artifact URI. From the same GitHub page, "Code scanning interprets results that are reported with relative paths as relative to the root of the repository analyzed. If a result contains an absolute URI, the URI is converted to a relative URI." A path that does not resolve to a committed file does not become a rendered alert, and in the worst case the whole upload is refused. `aquasecurity/trivy-action#471` reports exactly that shape, with the upload rejected carrying `locationFromSarifResult: expected artifact location`. `check` with no argument reads the newest capture out of the state directory, which on a runner is never inside the checkout.

Second, the frame sequence number is not a line number. SARIF 2.1.0 §3.30.5 says a region "MAY contain a property named startLine whose value is the one-based line number of the first character of the region" (OASIS, SARIF v2.1.0). `Seq` is scoped per session and it skips over dropped frames, so it diverges from the line index in two cases `check` already supports. Both show up in one run over a two-session log whose first session is missing `Seq` 3 and 4.

```
$ mcpsnoop check --fail-on error,invalid,warn,pending,incomplete two.jsonl
session alpha-1-aaaa: errors=0 invalid=0 warnings=1 mismatches=0 pending=0 deprecated=0 missing_frames=2
check failed: warn,incomplete
session beta-2-bbbb: errors=0 invalid=0 warnings=0 mismatches=0 pending=2 deprecated=0 missing_frames=0
check failed: pending
```

In that file line 3 holds `Seq` 5 and line 4 holds `Seq` 1 of a different session. Writing `Seq` into `startLine` would send a reader to the wrong frame, which is worse than sending them nowhere.

Suggested rule, to confirm before anyone implements.

- `artifactLocation.uri` is the session log path, made relative to the process working directory when the log sits under it, and left as an absolute `file:` URI otherwise.
- `region.startLine` is the real 1-based line the envelope was decoded from, counted during load rather than taken from `Seq`. `internal/exporter/exporter.go:266` decodes with a `json.Decoder` and keeps no offset, so this wants a newline-counting reader or `Decoder.InputOffset`.
- The session id and the frame `Seq` go into the message text and into `partialFingerprints`, never into the region.
- README states plainly that alerts render only when the log sits inside the checkout. The existing recipe already satisfies that, since it checks `artifacts/session.jsonl`.
- `check -` has no path at all, so a stdin run emits results with no `physicalLocation`.

The alternative of dropping `physicalLocation` everywhere and emitting `logicalLocations` only is legal SARIF but GitHub will not render it, which defeats the point of the format.

## Proposed change

Add `sarif` alongside `junit`, written by a new `cmd/mcpsnoop/check_sarif.go` that is a sibling of `cmd/mcpsnoop/check_junit.go`.

One rule per signal, ids `mcpsnoop/error`, `mcpsnoop/invalid`, `mcpsnoop/warn`, `mcpsnoop/mismatch`, `mcpsnoop/pending`, `mcpsnoop/drift`, `mcpsnoop/deprecated`, `mcpsnoop/incomplete`, plus `mcpsnoop/assertion` for the `--max-duration`, `--expect-tool` and `--forbid-tool` failures that `checkAssertions.eval` already produces (`cmd/mcpsnoop/check.go:162`).

One result per finding rather than per session. Every anchor below already exists on the view types.

| Rule | Where the finding comes from | Frame it anchors to |
|---|---|---|
| `error` | `CallView.Errored` (`internal/store/views.go:29`) | the response frame |
| `invalid` | `EventView.Kind == store.EventInvalid`, counted at `cmd/mcpsnoop/check.go:274` | the frame itself |
| `warn` | `EventView.Warning` (`internal/store/views.go:67`), counted at `cmd/mcpsnoop/check.go:277` | the frame itself |
| `mismatch` | `EventView.RoutingMismatch` (`internal/store/views.go:90`), counted at `cmd/mcpsnoop/check.go:280` | the request frame |
| `deprecated` | `EventView.Deprecated` (`internal/store/views.go:96`), counted at `cmd/mcpsnoop/check.go:283` | the frame itself |
| `pending` | `CallView.State`, reached through `EventView.Call` (`internal/store/views.go:102`) | the request frame that never got an answer |
| `drift` | `store.ToolDrift.Changes` (`internal/store/views.go:250-252`) | the `tools/list` response frame |
| `incomplete` | `SessionHeader.MissingFrames` (`internal/store/views.go:127`) | no frame survives, so no region |

Level follows `--fail-on` so the gate and the report never disagree. A signal inside the selection is `error`, a signal outside it is `note`, which surfaces the observation without asserting anyone must act on it.

One run per file rather than one per session. GitHub caps a SARIF file at 20 runs, and a concatenated capture can hold more sessions than that, so the session id belongs on each result instead.

Exit codes stay exactly as the junit path has them, report to stdout then `exitCode(1)` on failure (`cmd/mcpsnoop/check.go:95-102`), so the workflow step needs `continue-on-error` and the upload step needs `if: always()`.

One piece of plumbing. `loadCheckSession` (`cmd/mcpsnoop/check.go:242-251`) resolves the log path through `exporter.ResolveSessionPath` and then discards it, returning only the store and the first session id. SARIF needs that path, so it has to come back out.

## Acceptance criteria

- [ ] `mcpsnoop check --format sarif` writes a SARIF 2.1.0 log to stdout, and the `--format` help text at `cmd/mcpsnoop/check.go:142` lists all three formats.
- [ ] `parseCheckOutputFormat` accepts `sarif` and its error message names all three formats (`cmd/mcpsnoop/check.go:231-240`).
- [ ] Output validates against the SARIF 2.1.0 schema, with `$schema`, `version`, `runs[].tool.driver.name`, `runs[].tool.driver.version` from the build version, and `runs[].tool.driver.informationUri`.
- [ ] Exactly one rule per signal plus one for assertions, each with `id`, `name`, `shortDescription` and `fullDescription`, and every rule present even when the run is clean, so a fixed finding closes its alert instead of staying open forever.
- [ ] One result per finding, carrying the session id, the frame `Seq` and the warning or drift text, not a per-session count. The three warnings in the Context section come out as three results with three distinct messages.
- [ ] `level` is `error` for a signal named in `--fail-on` and `note` for one that is not, covered by a test that flips the selection and asserts the levels move with it.
- [ ] `region.startLine` is the true 1-based line the envelope was decoded from, proved by a test over a two-session log with a `Seq` gap where line and `Seq` differ for every frame after the gap.
- [ ] `partialFingerprints` is stable across two runs over the same log, derived from the session id, the rule id and the frame identity rather than from the line number.
- [ ] `incomplete` emits one result per affected session with no `region`, since a dropped frame leaves no line to point at.
- [ ] The first-seen tool baseline appears the way the junit writer marks it with `<skipped>` (`cmd/mcpsnoop/check_junit.go:81-89`), so a run that verified nothing never reads as clean.
- [ ] A clean session emits a valid log with an empty `results` array and exits 0. Exit codes otherwise match the junit path, 1 on a failing signal or assertion and 2 on a bad flag.
- [ ] Assertion failures from `--max-duration`, `--expect-tool` and `--forbid-tool` appear as results under `mcpsnoop/assertion`.
- [ ] `check -` emits results without `physicalLocation` rather than inventing a path.
- [ ] README gains `sarif` in the `--format` synopsis at `README.md:363`, a paragraph next to the junit one at `README.md:394`, a workflow snippet using `github/codeql-action/upload-sarif` with `if: always()`, and a sentence saying alerts render only when the log sits inside the checkout. The junit example stays.
- [ ] `make check` is green and `cmd/mcpsnoop/check_sarif_test.go` covers the clean run, the multi-session run, the `Seq` gap, the level mapping, and the stdin case.

## Out of scope

- Uploading anything. mcpsnoop makes no network calls of its own and this must not be the first one. The upload stays a step in the user's workflow.
- New signals, new conformance rules, or any new analysis. This is a serializer over what `summarizeCheck` (`cmd/mcpsnoop/check.go:256`) and `checkAssertions.eval` already return.
- SARIF for `export`. `export` describes a session, `check` produces findings, and only findings belong in SARIF.
- `fixes[]`, `codeFlows` and `taxonomies`. Nothing here suggests an edit to make, and there is no taxonomy to map to.
- Any consumer's own SARIF dialect beyond plain 2.1.0, including GitLab and Bitbucket variants.
- Changing what `--fail-on` gates or what any signal means. The report follows the gate, never the other way round.

## Files

- `cmd/mcpsnoop/check.go`, add `checkFormatSARIF` beside `:33-34`, extend `parseCheckOutputFormat` at `:231-240`, branch alongside the junit path at `:95-102`, update the flag help at `:142`, and return the resolved log path from `loadCheckSession` at `:242-251`.
- `cmd/mcpsnoop/check_sarif.go`, new, modelled on `cmd/mcpsnoop/check_junit.go`.
- `cmd/mcpsnoop/check_sarif_test.go`, new.
- `internal/exporter/exporter.go:266`, track the line each envelope was decoded from so a result can carry a real `startLine`.
- `internal/store/views.go`, read only. `EventView.Seq` at `:59`, `Warning` at `:67`, `RoutingMismatch` at `:90`, `Deprecated` at `:96`, `Call` at `:102`, `SessionHeader.MissingFrames` at `:127`, `ToolDrift` at `:250-260`.
- `README.md:355-427`, the CI section.

hsdfat:
Closes #205.

`check` emitted text and JUnit but nothing code scanning could ingest, so findings could fail a build without ever becoming an alert. Adds `--format sarif` (SARIF 2.1.0).

- one run per file, not per session — code scanning caps a file at 20 runs, and a concatenated capture can hold more
- every rule declared on every run, so a fixed finding's alert closes
- level follows `--fail-on`; a clean run emits `"results": []`, not `null`
- results anchor at the true line, but fingerprints are built from what a finding *is*, so alerts survive lines moving

One design note worth a look: the `error` signal comes from a new `EventView.Errored`, set in the store wherever the session's error counter is incremented. Deriving it from `Call.Errored` instead was wrong — some counted errors carry no call at all (a bare HTTP failure, an error response matching no request), and a task's terminal failure mutates its call in place, which anchored the alert at the call's *first* response. Marking the frame where it's counted keeps the report and the gate in sync by construction.

README gains the `--format` synopsis and an `upload-sarif` snippet.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,7 +360,7 @@ leave the capture incomplete, tool-definition drift, or use of deprecated
 protocol features.
 
 ```bash
-mcpsnoop check [--format text|junit] [--fail-on error,invalid,warn,mismatch,pending,drift,deprecated,incomplete] [session-id|log.jsonl|-]
+mcpsnoop check [--format text|junit|sarif] [--fail-on error,invalid,warn,mismatch,pending,drift,deprecated,incomplete] [session-id|log.jsonl|-]
 ```
 
 The three default signals (error, invalid, warn) fail the check. Add `pending`
@@ -393,6 +393,15 @@ definition remain observational only. Existing key- and value-based redaction
 also applies to captured parameter-header values before they reach a sink.
 Use `--format junit` to write one JUnit `<testcase>` per signal and session;
 failures follow the same `--fail-on` selection as the text output.
+Use `--format sarif` to write a SARIF 2.1.0 log instead. Where junit reports one
+aggregate per signal, SARIF reports one result per finding, carrying the session,
+the frame `Seq` and the frame's own warning or drift text, and pointing at the
+line of the log the frame was decoded from. A signal named in `--fail-on` is
+reported at level `error` and one outside it at level `note`, so the report and
+the gate never disagree. Alerts render only when the session log sits inside the
+checkout, since code scanning resolves a result's path against the repository
+root; the `artifacts/session.jsonl` recipe below already satisfies that, while a
+log read from the state directory or from stdin does not.
 
 ```bash
 mcpsnoop check build-agent
@@ -425,6 +434,23 @@ mcpsnoop check --expect-tool search --forbid-tool delete --max-duration 2s run.j
   with:
     name: mcpsnoop-junit
     path: test-results/mcpsnoop.xml
+```
+
+To put the findings in the Security tab instead, hand the SARIF log to
+`upload-sarif`. `check` exits non-zero on a finding, so the check step needs
+`continue-on-error` and the upload needs `if: always()`, or the report never
+reaches the tab on the runs that have something to report.
+
+```yaml
+- name: Check captured MCP session
+  continue-on-error: true
+  run: mcpsnoop check --format sarif artifacts/session.jsonl > mcpsnoop.sarif
+- name: Upload mcpsnoop SARIF report
+  if: always()
+  uses: github/codeql-action/upload-sarif@v3
+  with:
+    sarif_file: mcpsnoop.sarif
+    category: mcpsnoop
 ```
 
 ### Detect tool definition drift
--- a/README.md
+++ b/README.md
@@ -402,10 +402,19 @@ aggregate per signal, SARIF reports one result per finding, carrying the session
 the frame `Seq` and the frame's own warning or drift text, and pointing at the
 line of the log the frame was decoded from. A signal named in `--fail-on` is
 reported at level `error` and one outside it at level `note`, so the report and
-the gate never disagree. Alerts render only when the session log sits inside the
-checkout, since code scanning resolves a result's path against the repository
-root; the `artifacts/session.jsonl` recipe below already satisfies that, while a
-log read from the state directory or from stdin does not.
+the gate never disagree. Code scanning rejects a file whose run holds more than
+25,000 results and displays only the top 5,000 of what it accepts, so the report
+is capped at 5,000: the findings the gate failed on first, then a
+`mcpsnoop/report-truncated` result saying how many were left out. The text and
+junit formats stay complete.
+
+A result points at the log with a path relative to the working directory, which
+code scanning then resolves against the repository root. The alert renders with
+its surrounding lines only when that path is a file in the analysed commit, so a
+capture the workflow generated into `artifacts/` opens an alert carrying the
+message, the rule and the line number but no source view. Committing a capture
+you want rendered in full is the only way to get one. A log read from the state
+directory or from stdin gets no path at all.
 
 ```bash
 mcpsnoop check build-agent
@@ -441,17 +450,28 @@ mcpsnoop check --expect-tool search --forbid-tool delete --max-duration 2s run.j
 ```
 
 To put the findings in the Security tab instead, hand the SARIF log to
-`upload-sarif`. `check` exits non-zero on a finding, so the check step needs
-`continue-on-error` and the upload needs `if: always()`, or the report never
-reaches the tab on the runs that have something to report.
+`upload-sarif`. The job needs `security-events: write`, or the upload answers
+403. `check` exits non-zero on a finding, so the upload step needs `if: always()`
+to run at all on the runs that have something to report; `continue-on-error`
+hands the verdict to the code scanning check, which fails on an `error`-level
+alert and can be made a required check. Drop it if you would rather the check
+step itself be what turns the job red.
 
 ```yaml
+permissions:
+  # required for all workflows
+  security-events: write
+  # only required for workflows in private repositories
+  actions: read
+  contents: read
+
+steps:
 - name: Check captured MCP session
   continue-on-error: true
   run: mcpsnoop check --format sarif artifacts/session.jsonl > mcpsnoop.sarif
 - name: Upload mcpsnoop SARIF report
   if: always()
-  uses: github/codeql-action/upload-sarif@v3
+  uses: github/codeql-action/upload-sarif@v4
   with:
     sarif_file: mcpsnoop.sarif
     category: mcpsnoop
--- a/cmd/mcpsnoop/baseline.go
+++ b/cmd/mcpsnoop/baseline.go
@@ -39,11 +39,12 @@ func newBaselineCmd() *cobra.Command {
 			if len(args) == 1 {
 				arg = args[0]
 			}
-			st, sessionID, err := loadCheckSession(cmd, arg)
+			sessionLog, err := loadCheckSession(cmd, arg)
 			if err != nil {
 				fmt.Fprintln(cmd.ErrOrStderr(), "mcpsnoop baseline:", err)
 				return exitCode(1)
 			}
+			st, sessionID := sessionLog.store, sessionLog.sessionID
 			manager := toolbaseline.New(resolveBaselineDir(baselineDir))
 			switch {
 			case accept:
--- a/cmd/mcpsnoop/baseline.go
+++ b/cmd/mcpsnoop/baseline.go
@@ -39,7 +39,8 @@ func newBaselineCmd() *cobra.Command {
 			if len(args) == 1 {
 				arg = args[0]
 			}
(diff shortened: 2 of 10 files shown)


kerlenton:
Great! Thank you!

PR #213, opened 2026-08-02
feat(cli): wrap and unwrap the Claude Desktop config for one server

### LINKED ISSUE #136: mcpsnoop wrap for Claude Desktop, so the config edit stops being manual
## Context

Everything about mcpsnoop is zero-config except the one step at the very start, hand-editing a client config to wrap the server command. It is the only manual operation left, and it lands before the user has any reason to trust the tool enough to edit a config for it.

This stalled before because config formats differ across clients and drift over time, which made the how unclear. Narrowing to one client removes that objection without pretending the wider problem is solved.

## Proposed change

`wrap` locates the config, backs it up, and rewrites the named server's command to run through `mcpsnoop --`. `unwrap` restores it. Re-running `wrap` does not double-wrap. `--dry-run` prints the change without writing. A missing config or an unknown server name is a clear message, not a stack trace.

Do Claude Desktop completely, and structure it so a second client is a new file rather than a rewrite of this one.

## Scope boundary

One client. No discovery across clients, no touching a config the user did not name, and nothing written without a backup first.

## Acceptance criteria

- [ ] `wrap` edits the config and `unwrap` restores it byte for byte
- [ ] Idempotent under repeated `wrap`
- [ ] `--dry-run` writes nothing
- [ ] Clear errors for a missing config or an unknown server
- [ ] Adding a second client later means adding a file, not changing this one
- [ ] Tests against fixture configs for all of the above

## Files

- new `cmd/mcpsnoop/wrap.go`, registered in `cmd/mcpsnoop/main.go`
- `internal/paths/paths.go`

hsdfat:
Closes #136.

Pointing Claude Desktop at the proxy meant hand-editing its config — easy to get wrong, awkward to undo. Adds `mcpsnoop wrap` / `unwrap` for one named server.

- backs the config up before touching it, and writes atomically (same pattern as `internal/toolbaseline/baseline.go`)
- wrapping twice is a no-op rather than nesting; `unwrap` restores the entry byte for byte
- `--dry-run` prints the change without writing
- an unknown server or missing config is an error, never a silent rewrite

Scoped to Claude Desktop only, one config path per OS, as the issue asks. Client-specific details sit behind a small `wrapClient` value so a second client would be a data change — but none is added here.

## Checklist

- [x] `make check` passes (gofmt, vet, staticcheck, race tests) — plus cross-compiled vet and build for windows and darwin

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,10 +58,12 @@ mcpsnoop unwrap my-server   # put it back
 ```
 
 `wrap` finds `claude_desktop_config.json`, copies it to
-`claude_desktop_config.json.mcpsnoop.bak`, and rewrites only that one server's
-entry, so your formatting, key order and every other server are left alone.
-`unwrap` restores the file. Restart Claude Desktop after either, since MCP
-servers are launched once at startup.
+`claude_desktop_config.json.mcpsnoop.bak` the first time, and rewrites only that
+one server's entry, so your formatting and every other server are left alone.
+Inside the rewritten entry the keys come back in alphabetical order. `unwrap`
+restores the file, and removes the backup once no server is wrapped any more.
+Restart Claude Desktop after either, since MCP servers are launched once at
+startup.
 
 Then use your client as usual and open the UI.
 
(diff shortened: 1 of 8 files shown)


kerlenton:
Thank you!

PR #215, opened 2026-08-07
fix(store): report non-object inputSchema and surface schema findings

### LINKED ISSUE #199: A tool inputSchema that is not an object schema is never reported, and no schema finding reaches check or export
## Context

A tool's `inputSchema` is the one part of a `tools/list` response that a client validates before it will register the tool, and a schema that fails that validation fails silently on the wire. The server answers `initialize`, returns a well-formed `tools/list`, and holds the connection open, and no `tools/call` ever arrives. Nothing in the traffic says why, which is why this class of defect gets measured in weeks.

[modelcontextprotocol/servers#3051](https://github.com/modelcontextprotocol/servers/issues/3051) is the canonical case. `@modelcontextprotocol/server-filesystem` shipped with every tool's `inputSchema` reduced to a dialect marker and nothing else.

```json
{
  "name": "read_file",
  "description": "Read the complete contents of a file",
  "inputSchema": {
    "$schema": "http://json-schema.org/draft-07/schema#"
  }
}
```

What the reporter saw was the SDK rejecting all of them at once, the same error repeated per tool index.

```
[
  {
    "code": "invalid_literal",
    "expected": "object",
    "path": [
      "tools",
      0,
      "inputSchema",
      "type"
    ],
    "message": "Invalid literal value, expected \"object\""
  },
```

Twenty-three comments follow, most of them bisecting npm tarballs. From the reporter partway through, "I did go back in time and installed all versions of the fileserver and somehow no version worked." A third party named the field days later. "In the August 23rd version, the `inputSchema` was generated via the `zodToJsonSchema` method from "zod-to-json-schema" package. However, it appears this method was used incorrectly, resulting in all tool's `inputSchemas` being `{"$schema":"http://json-schema.org/draft-07/schema#"}`. and failed validation by the MCP SDK."

[modelcontextprotocol/servers#4545](https://github.com/modelcontextprotocol/servers/issues/4545) is the same class one client further along, and it states the wire signature outright. "`mcp-server-Filesystem.log` shows every session since the update doing a successful `initialize` + `tools/list`, then **zero `tools/call` messages ever arriving**." A macOS reporter narrowed the registration-time variant to the schema bytes. "2025.8.21 stamps $schema draft-07 into inputSchema too (visible in its raw tools/list) ... macOS Desktop apparently validates at registration and rejects on $schema alone, hence 'no tools available'". The thread's working fix was a hand-rolled stdio shim that scrubs `$schema`, `outputSchema` and `annotations` out of `tools/list`, which is mcpsnoop's job done blind. That thread later narrowed away from `$schema` for the older package generation, and the client shipped a fix in 1.24012.9, so the claim here is not that `$schema` is illegal. The claim is that four fields got scrubbed at once because nothing in the path would name one.

Maintainer side, [modelcontextprotocol#2806](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2806) asks for tooling of exactly this shape. @olaservo writes "we should put something together to cover the mismatch between what LLM providers actually support vs what the spec allows.  It might not belong in the spec itself, but there should be more details on this somewhere adjacent to the spec, so that implementers don't run into it through trial and error."

### What mcpsnoop does with that traffic today

Built from `main` at `becf38e` and wrapped around a stdio server that advertises three tools. `read_file` carries the servers#3051 shape above, `write_file` carries the servers#4545 shape (a real object schema with a draft-07 `$schema` stamped in), and `get_time` carries the spec's own recommended no-parameter shape. The listing declares `ttlMs` and `cacheScope`, so nothing unrelated fires.

```
$ mcpsnoop check schemarepro-99913-791bf1aac5e7
session schemarepro-99913-791bf1aac5e7: errors=0 invalid=0 warnings=0 mismatches=0 pending=0 deprecated=0 missing_frames=0
recorded first-seen tool baseline (trusted, not verified)
check passed
$ echo "exit status $?"
exit status 0
```

Selecting every signal changes nothing.

```
$ mcpsnoop check schemarepro-99913-791bf1aac5e7 --fail-on error,invalid,warn,mismatch,pending,drift,deprecated,incomplete
session schemarepro-99913-791bf1aac5e7: errors=0 invalid=0 warnings=0 mismatches=0 pending=0 deprecated=0 missing_frames=0
check passed
$ echo "exit status $?"
exit status 0
```

A server whose every tool is unusable in an SDK that validates gets a clean bill of health. Two independent gaps produce that green.

**The schema walk never reads the root.** `analyzeSchema` at `internal/store/schema.go:35` decodes the schema and hands it to `walkSchema` at `internal/store/schema.go:77`, which looks only for composition keywords, `$ref`, and untyped entries under `properties`. The root `type` is read once, at `internal/store/schema.go:66`, by `isTyped`, and only for a child of `properties`. `$schema` is not read anywhere in the file. Calling `analyzeSchema` directly over the shapes involved gives this.

```
servers#3051 shape           {"$schema":"http://json-schema.org/draft-07/schema#"} -> []
servers#4545 shape           {"$schema":"http://json-schema.org/draft-07/schema#","type":"object","properties":{"path":{"type":"string"}}} -> []
array root                   {"type":"array","items":{"type":"string"}} -> []
empty object                 {} -> []
boolean schema true          true -> []
null schema                  null -> []
spec recommended no-param    {"type":"object","additionalProperties":false} -> []
oneOf, for contrast          {"type":"object","properties":{"a":{"oneOf":[{"type":"string"}]}}} -> [{oneOf}]
```

**No finding reaches `check` or `export` even when one exists.** `SchemaFinding` has exactly one consumer. `internal/tui/view.go:1413` builds a per-name map and `internal/tui/view.go:1521` paints it into the SCHEMA column. Adding a fourth tool whose schema does produce a finding today, a property with `oneOf`, and then enumerating every key of that session's `mcpsnoop export --format json` yields only `$schema`, `inputSchema` and `schema_bytes`. The first two are raw copied frame bytes and the third is a byte count. Nothing computed crosses over, and the per-tool block a reader would look in carries costs alone.

```
$ mcpsnoop export <session> --format json     # summary.definitions.per_tool[0]
{
  "name": "write_file",
  "bytes": 235,
  "description_bytes": 17,
  "schema_bytes": 167
}
```

`grep -rn Findings cmd internal/exporter` returns nothing.

Drift is the closest thing that fires, and only against a trusted baseline. Baselining a good `write_file` and then capturing the servers#4545 regression gives the tool name with no reason attached.

```
$ mcpsnoop check schemarepro-99870-f5005b2ef962 --fail-on drift
session schemarepro-99870-f5005b2ef962: errors=0 invalid=0 warnings=1 mismatches=0 pending=0 deprecated=0 missing_frames=0
definition drift:
  input schema changed: write_file
check failed: drift
```

A first capture of an already broken server, which is the servers#3051 situation, has no baseline and gets nothing at all.

### Normative basis

From `schema/2026-07-28/schema.json`, `$defs.Tool.properties.inputSchema`, verbatim.

```json
{
  "additionalProperties": {},
  "description": "A JSON Schema object defining the expected parameters for the tool.\n\nTool arguments are always JSON objects, so `type: \"object\"` is required at the root. ...",
  "properties": {
    "$schema": { "type": "string" },
    "type": { "const": "object", "type": "string" }
  },
  "required": [
    "type"
  ],
  "type": "object"
}
```

`$defs.Tool.required` is `["inputSchema", "name"]`, so an absent `inputSchema` fails the same way a rootless one does. The identical `"required": ["type"]` with `"type": {"const": "object"}` appears in `schema/2025-11-25/schema.json` and in `schema/2025-06-18/schema.json` under `definitions`, so this needs no revision gate.

The prose agrees. [Tools](https://modelcontextprotocol.io/specification/2026-07-28/server/tools) says `inputSchema` "**MUST** be a valid JSON Schema object (not `null`)" and gives the only two acceptable no-parameter forms as `{ "type": "object", "additionalProperties": false }` and `{ "type": "object" }`.

`outputSchema` carries no such constraint. Its entry in the same schema file has no `required` and no `const`, and the Tools page shows an array-rooted `outputSchema` in the `list_users` example, so the root-type rule is an `inputSchema` rule only.

The dialect is a separate question and it is legal. The Tools page says `inputSchema` "Defaults to 2020-12 if no `$schema` field is present" and shows a draft-07 `inputSchema` under Schema Examples. [Overview](https://modelcontextprotocol.io/specification/2026-07-28/basic/index) is explicit that "Schemas MAY include a `$schema` field to specify a different dialect", while requiring implementations to "support at least 2020-12" and to "handle unsupported dialects gracefully by returning an appropriate error indicating the dialect is not supported". A non-2020-12 dialect therefore belongs in the observation channel and never in the warning channel.

## Decision needed

Which channel the root-type violation rides, since a warning fails a default `check` run.

The case for `ev.warning`, the channel the existing per-tool schema problem already uses at `internal/store/store.go:436`, is that the condition is decidable from the wire with nothing guessed, the constraint holds in every revision mcpsnoop can observe, and the consequence for the user is total loss of the tool. The cost is that a CI job green today against a server that was always broken turns red on upgrade. That is the right answer for this defect and it is still a behaviour change that belongs in the release notes.

The alternative is a structured field that never fails by default and is opt-in through `--fail-on`, which is how `Deprecated` works at `internal/store/views.go:94-96`.

The rest of this issue assumes the first for the violation and the second for the dialect observation. Say so on the issue before implementing if that split is wrong.

## Proposed change

Three parts, all purely syntactic.

**1. Report an `inputSchema` that is not an object schema.** Add a finding kind, `nonObjectRoot`, set when a tool's `inputSchema` is absent, is not a JSON object, or is an object whose `type` is absent or is any value other than the string `"object"`. Nothing is resolved and no dialect is interpreted. Attach it as a warning on the `tools/list` response frame in the same place the `x-mcp-header` violation is attached at `internal/store/store.go:436`, naming the tool and the root type seen. Attaching at ingest also means it survives a partly paginated listing, which a finding carried only on `ToolDefinition` does not, since `ToolDefinitions` at `internal/store/views.go:438` returns nothing until `toolListComplete`.

**2. Record a non-default dialect as an observation.** Add a second kind, `nonDefaultDialect`, set when `inputSchema` or `outputSchema` carries a `$schema` whose value is not the 2020-12 URI. Never a warning, because the spec permits it. This is the field the servers#4545 shim scrubbed blind.

**3. Surface findings in `check` and `export`.** Give `ToolCostExport` a findings list keyed by the kind string, print a schema section in `check` text output shaped like `writeToolDrift`, and give the observational kinds an opt-in `--fail-on` signal. Keyed structurally, never on display text, the same rule `RoutingMismatch` follows at `internal/store/views.go:82-90`. This also makes drift explicable, since "input schema changed" plus "read_file now has no root type" is a diagnosis and the first line alone is not.

## Acceptance criteria

- [ ] `analyzeSchema` reports `nonObjectRoot` for `{"$schema":"http://json-schema.org/draft-07/schema#"}`, `{}`, `{"type":"array","items":{"type":"string"}}`, `true`, `null`, and an `inputSchema` that is absent from the tool entirely.
- [ ] `analyzeSchema` reports nothing for `{"type":"object"}` or `{"type":"object","additionalProperties":false}`, the two forms the Tools page names for a tool with no parameters.
- [ ] `nonObjectRoot` is never reported for `outputSchema`, and the array-rooted `list_users` `outputSchema` from the Tools page stays silent.
- [ ] `nonDefaultDialect` is reported for a draft-07 `$schema` on either `inputSchema` or `outputSchema`, and is not reported when `$schema` is absent or is the 2020-12 URI.
- [ ] The `tools/list` response frame carries a warning naming each offending tool, so `mcpsnoop check` on the servers#3051 capture above exits non-zero with `read_file` named in the output.
- [ ] The dialect observation never fails a default `check` run, and the servers#4545 capture stays green until its signal is selected.
- [ ] `mcpsnoop export --format json` carries per-tool findings keyed by kind string, and the html and text writers show them.
- [ ] The findings that already exist (`oneOf`, `anyOf`, `allOf`, `not`, `ref`, `externalRef`, `untypedProperty`) reach `check` and `export` too, not only the two new ones.
- [ ] A `tools/list` that never finished paginating still reports the violation.
- [ ] `schemaHeadlineOrder` in the TUI ranks `nonObjectRoot` above every existing kind, since it is the only one that is a violation.
- [ ] `mcpsnoop help check`, the `--fail-on` flag text, and `docs/` describe any new signal.

## Files

- `internal/store/schema.go`, kinds at 13-21, `analyzeSchema` at 35, `isTyped` at 65, `walkSchema` at 77. The root `type` and `$schema` are both unread today.
- `internal/store/store.go`, `inputSchema` decoded at 1150, `analyzeSchema` called at 1211, and 436-441 as the precedent for turning a per-tool schema problem into a warning on the `tools/list` response.
- `internal/store/views.go`, `ToolDefinition.Findings` at 145, `ToolDefinitions` at 438, `Event.Deprecated` at 94-96 as the model for a marker that never fails by default.
- `internal/exporter/exporter.go`, `ToolListCostExport` at 56 and `ToolCostExport` at 65, where per-tool findings would land.
- `cmd/mcpsnoop/check.go`, the `--fail-on` flag at 141, signal parsing at 225, `summarizeCheck` at 256, and the per-session output block at 104-121.
- `cmd/mcpsnoop/baseline.go`, `writeToolDrift` at 94, the shape a schema section should match.
- `internal/tui/view.go`, 1413-1418 and 1521 and 1572-1607, which is the whole of today's consumption.

## Out of scope

- Validating a schema against its declared dialect. That needs a JSON Schema implementation, and the verdict stops being decidable by inspecting the frame.
- Resolving or fetching a `$ref`. The spec forbids automatic dereferencing of network `$ref`s, and mcpsnoop makes no network calls of its own.
- Any prediction about whether a particular client will accept a particular schema. The dialect finding says what the bytes are, not what a client does with them.
- Rewriting or scrubbing frames the way the servers#4545 shim does. mcpsnoop observes and the shim stays dumb.
- Root-type rules for `outputSchema`. There are none to check.

syf2211:
## Summary

Detect and report tool `inputSchema` values that are not valid object schemas (per MCP spec), record non-default JSON Schema dialects as observations, and surface all schema findings through `check`, `export`, and an opt-in `--fail-on schema` signal.

## Motivation

Fixes #199 — servers like `@modelcontextprotocol/server-filesystem` can advertise tools whose `inputSchema` is only `{"":"..."}` with no `type: object`. SDKs reject these silently, but `mcpsnoop check` previously passed with zero warnings. Schema composition findings (`oneOf`, etc.) were computed but only shown in the TUI.

## Changes

- Add `nonObjectRoot` finding for absent/invalid inputSchema roots; warn on the `tools/list` response frame
- Add `nonDefaultDialect` observation for non-2020-12 `$schema` on inputSchema and outputSchema
- Merge outputSchema dialect findings into per-tool `Findings`
- Print `schema findings:` section in `check` text output; add `schema` to `--fail-on`
- Include per-tool `findings` in JSON/text/HTML export
- Rank `nonObjectRoot` first in TUI schema headline order

## Tests

```
go test ./internal/store ./cmd/mcpsnoop ./internal/exporter
```

All pass. Full suite except two pre-existing TUI style tests on upstream main (`TestStreamRowShowsSupersededStatusInWarnStyle`, `TestStatusStyleWarnsOnTruncated`).

## Notes

- `nonObjectRoot` fails default `check` via the existing `warn` signal (warning on frame)
- Dialect/composition findings require `--fail-on schema` to fail CI
- Reviewed with composer-2.5 subagent before submission

### DIFF
--- a/README.md
+++ b/README.md
@@ -570,21 +570,37 @@ than one kind.
 | Shown | Means |
 |---|---|
 | `no root` | the `inputSchema` is absent, is not a JSON object, or has a root type other than `"object"` |
+| `dialect` | a `$schema` naming a dialect other than the 2020-12 the revision defaults to |
 | `ext ref` | a `$ref` pointing outside the document, which is also the case the spec warns implementers not to follow blindly |
 | `oneOf`, `anyOf`, `allOf`, `not` | a composition keyword, handled inconsistently across clients |
 | `ref` | a `$ref` pointing inside the same document |
 | `untyped` | a property that declares no type and no other way of saying what it accepts |
 
 All but the first are observations rather than verdicts. A schema using `oneOf`
-is not wrong, only likely to be read differently by different clients. `no root`
-is the exception: the `Tool` definition requires `inputSchema` and pins its root
-type to `"object"`, so a client validating a listing rejects that tool outright
-and it never becomes callable, with nothing on the wire to say why. `no root`
-leads the column for that reason, and a schema mcpsnoop's own redaction scrubbed
-is never reported, since an unreadable schema is not a wrong one.
+is not wrong, only likely to be read differently by different clients, and a
+schema may declare whatever dialect it likes. `no root` is the exception: the
+`Tool` definition requires `inputSchema` and pins its root type to `"object"`, so
+a client validating a listing rejects that tool outright and it never becomes
+callable, with nothing on the wire to say why. `no root` leads the column for
+that reason, and a schema mcpsnoop's own redaction scrubbed is never reported,
+since an unreadable schema is not a wrong one.
 
-The column carries the warning color and never the red of the ERR column. There
-is no `check` signal for any of it yet, and nothing about MCP traffic changes.
+That split decides what `check` does with them. `no root` is a warning on the
+`tools/list` frame, so it fails the default `error,invalid,warn` gate with no
+flag at all, which is the point: a server that ships an unusable tool answers
+every handshake normally and simply never receives a `tools/call`. The
+observations are counted as `schema_findings` and reported under `schema
+findings:`, and only fail the run when you add `schema` to `--fail-on`. Both
+reach `--format junit` and `--format sarif`, and `export` carries the per-tool
+list under `summary.definitions.per_tool[].findings`.
+
+```bash
+mcpsnoop check session.jsonl                     # a non-object root already fails this
+mcpsnoop check --fail-on schema session.jsonl    # and now so do the observations
+```
+
+The column carries the warning color and never the red of the ERR column, and
+mcpsnoop still changes nothing about the traffic it forwards.
 
 Nothing is resolved or fetched. An external `$ref` is recognized by its form
 alone, and the schema it points at is never read.
--- a/README.md
+++ b/README.md
@@ -48,7 +48,24 @@ To use it for real, wrap your server in your client's MCP config.
 
 Everything after `--` is the command that normally launches your server. Swap in
 whatever you already use, like `python server.py`, `npx -y @scope/server`, or a
-compiled binary. Then use your client as usual and open the UI.
+compiled binary.
+
+On Claude Desktop you don't have to make that edit by hand.
+
+```bash
+mcpsnoop wrap my-server     # route my-server through mcpsnoop
+mcpsnoop unwrap my-server   # put it back
+```
+
+`wrap` finds `claude_desktop_config.json`, copies it to
+`claude_desktop_config.json.mcpsnoop.bak` the first time, and rewrites only that
+one server's entry, so your formatting and every other server are left alone.
+Inside the rewritten entry the keys come back in alphabetical order. `unwrap`
+restores the file, and removes the backup once no server is wrapped any more.
+Restart Claude Desktop after either, since MCP servers are launched once at
+startup.
+
+Then use your client as usual and open the UI.
 
 ```bash
 mcpsnoop
@@ -109,11 +126,13 @@ Explicit command-line flags override values from the config file.
 | `mcpsnoop` | open the live TUI |
 | `mcpsnoop http --target <url>` | proxy a streamable-HTTP server |
 | `mcpsnoop export` | render a session to json, html, text, har, or otlp |
-| `mcpsnoop check` | fail CI on errors, invalid frames, warnings, routing mismatches, or hung calls |
+| `mcpsnoop check` | fail CI on errors, invalid frames, warnings, routing mismatches, hung calls, or late results |
 | `mcpsnoop baseline` | inspect, accept, or reset trusted tool definitions |
 | `mcpsnoop diff` | compare tools and calls across two captured sessions |
 | `mcpsnoop open` | open a saved session in the TUI |
 | `mcpsnoop prune` | delete saved session logs older than a cutoff |
+| `mcpsnoop wrap <server>` | route one of Claude Desktop's servers through mcpsnoop |
+| `mcpsnoop unwrap <server>` | put that server's entry back the way it was |
 | `mcpsnoop remote <user@host>` | print the SSH tunnel command |
 | `mcpsnoop demo` | play a scripted session |
 
@@ -223,12 +242,14 @@ matches the method, tool, id, and payload.
 | `task:` | task id | `task:01J...` |
 | `dir:` | direction (`c2s`, `s2c`) | `dir:s2c` |
 | `kind:` | frame type (`req`, `resp`, `notify`, `stderr`, `invalid`) | `kind:invalid` |
-| `status:` | call outcome (`ok`, `error`, `cancelled`, `pending`, `bad`, `warn`, `mismatch`) | `status:error` |
+| `status:` | call outcome (`ok`, `error`, `cancel`, `late`, `cancelled`, `pending`, `bad`, `warn`, `mismatch`, or an HTTP status like `401`) | `status:error` |
 
 Stack tokens to get specific.
 
 ```text
 tool:search status:pending        # in-flight calls to one search tool
+status:cancel                     # calls the client gave up on (status:cancelled is a cancelled task)
+status:late                       # results that arrived after the cancellation
 method:tools/call status:error    # tool calls that failed
 dir:s2c kind:req                  # server-initiated requests (servers before 2026-07-28)
 ```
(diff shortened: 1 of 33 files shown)


kerlenton:
Thank you!

PR #216, opened 2026-08-08
fix(store): report a tool inputSchema that is not an object schema

DevMello:
Part 1 of #199.

A tool whose `inputSchema` has no root type is rejected by clients at
registration. `analyzeSchema` walked subschemas for constructs that travel
badly but never read the root type, so the one condition here that is a
violation was the one it could not report.

Adds a `nonObjectRoot` finding, set when `inputSchema` is absent, is not a JSON
object, or has a root type that is not `"object"`. Nothing is resolved and no
dialect is interpreted. The TUI ranks it above every existing kind, since the
others say a schema may be read differently across clients and this one says a
conforming client refuses the tool.

Nine existing fixtures were rootless schemas used to test construct detection,
so they trip the new finding too. I gave each a valid object root rather than
adding the new kind to every `want`, so each test keeps isolating one thing.

Reporting this on the `tools/list` frame so that `check` fails is a follow-up,
kept separate to keep this reviewable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -545,20 +545,27 @@ calling wrongly is often a tool whose schema asked for more than the client
 delivers.
 
 The tool summary, opened with `s`, has a SCHEMA column naming the most notable
-construct each advertised tool uses, with a trailing `+` when the schema uses
-more than one kind.
+thing about each advertised tool's schema, with a trailing `+` when there is more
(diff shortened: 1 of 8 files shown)


kerlenton:
LGTM! Thanks!

PR #219, opened 2026-08-10
feat(tui): edit replay request parameters

### LINKED ISSUE #207: Change an argument and re-send the call
## Context

`r` in the TUI replays the focused request. `replay.Replay` spawns a fresh isolated copy of the wrapped server, works out which revision it speaks, and sends one request built from a method and a `params` blob (`internal/replay/replay.go:47`). The TUI is its only caller, and it always hands over the captured params byte for byte (`internal/tui/model.go:804`, `internal/tui/replay.go:24`). Nothing lets you change them first, and there is no other send path in the binary. The `:` palette handles `q`, `help`, `sessions`, `stream`, `summary`, and `export`, and nothing else (`internal/tui/model.go:496`).

That matters because the reason to re-send a failed call is almost always to fix an argument. The 2026-07-28 Tools page frames tool execution errors as exactly that channel.

> **Tool Execution Errors** contain actionable feedback that language models can use to self-correct and retry with adjusted parameters

and lists among them

> Input validation errors (e.g., date in wrong format, value out of range)

(Tools, 2026-07-28, Error Handling.) A person in front of the TUI is doing that same self-correction by hand. mcpsnoop can only repeat the argument that already failed.

### Reproduction

A server with one tool that rejects a badly formatted date the way the spec describes.

```python
# srv.py
import sys, json, re
TOOLS = [{"name": "search_orders",
          "description": "Search orders by ISO date and region",
          "inputSchema": {"type": "object",
                          "properties": {"since": {"type": "string", "description": "ISO 8601 date, YYYY-MM-DD"},
                                         "region": {"type": "string"},
                                         "limit": {"type": "integer"}},
                          "required": ["since", "region"]}}]

def send(o):
    sys.stdout.write(json.dumps(o) + "\n"); sys.stdout.flush()

for line in sys.stdin:
    if not line.strip():
        continue
    m = json.loads(line)
    mid, meth = m.get("id"), m.get("method")
    if meth == "initialize":
        send({"jsonrpc":"2.0","id":mid,"result":{"protocolVersion":"2025-06-18",
              "capabilities":{"tools":{"listChanged":False}},
              "serverInfo":{"name":"orders","version":"1.0.0"}}})
    elif meth == "tools/list":
        send({"jsonrpc":"2.0","id":mid,"result":{"tools":TOOLS}})
    elif meth == "tools/call":
        args = m["params"].get("arguments") or {}
        since = args.get("since", "")
        if not re.match(r"^\d{4}-\d{2}-\d{2}$", str(since)):
            send({"jsonrpc":"2.0","id":mid,"result":{"content":[{"type":"text",
                  "text":"Invalid departure date: must be in the future. since=%r is not YYYY-MM-DD." % since}],
                  "isError":True}})
        else:
            send({"jsonrpc":"2.0","id":mid,"result":{"content":[{"type":"text",
                  "text":"3 orders since %s in %s" % (since, args.get("region"))}],
                  "isError":False}})
```

```sh
cat > client.sh <<'SH'
#!/bin/sh
printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"demo","version":"1"}}}'
sleep 0.3
printf '%s\n' '{"jsonrpc":"2.0","method":"notifications/initialized"}'
printf '%s\n' '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'
sleep 0.3
printf '%s\n' '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"search_orders","arguments":{"since":"last tuesday","region":"emea","limit":5}}}'
sleep 0.5
SH
chmod +x client.sh

./client.sh | mcpsnoop --trace-file ./sess.jsonl -- python3 ./srv.py >/dev/null
mcpsnoop open ./sess.jsonl
```

Enter the session and the failing call is right there.

```
╭─ stream · srv.py ───────────────────────────────────────────── 3 calls · p50 1ms · p95 22ms ─╮
│ TIME            METHOD / TOOL                    ID     DUR  STATUS   DETAIL                 │
│ 19:20:41.217  → initialize                        1                   {"protocolVersion":"2…│
│ 19:20:41.239  ← response                          1    22ms  ok       {"protocolVersion":"2…│
│ 19:20:41.522  · notify notifications/initialized                      {"jsonrpc":"2.0","me…│
│ 19:20:41.522  → tools/list                        2                   {}                     │
│ 19:20:41.522  ← response                          2      0s  ok       {"tools":[{"name":"se…│
│▌19:20:41.839  → tools/call search_orders          3                   {"name":"search_order…│
│ 19:20:41.840  ← response                          3     1ms  err      Invalid departure dat…│
╰──────────────────────────────────────────────────────────────────────────────────────────────╯

 enter inspect  r replay  c caps  s summary  / filter  p pause  ? help          7 frames · 1 err
```

Press `r` on the request, answer `y` to the command confirm, and this is what comes back.

```
╭──────────────────────────────────────────────────────────────────────────────────────────────╮
│  REPLAY · tools/call                                                                         │
│                                                                                              │
│  ok  (0s)                                                                                    │
│                                                                                              │
│  request params                                                                              │
│  {                                                                                           │
│    "name": "search_orders",                                                                  │
│    "arguments": {                                                                            │
│      "since": "last tuesday",                                                                │
│      "region": "emea",                                                                       │
│      "limit": 5                                                                              │
│    }                                                                                         │
│  }                                                                                           │
│                                                                                              │
│  response                                                                                    │
│  {                                                                                           │
│    "jsonrpc": "2.0",                                                                         │
│    "id": 2,                                                                                  │
│    "result": {                                                                               │
│      "content": [                                                                            │
│        {                                                                                     │
│          "type": "text",                                                                     │
│          "text": "Invalid departure date: must be in the future. since='last tuesday' is …"  │
│        }                                                                                     │
│      ],                                                                                      │
│      "isError": true                                                                         │
│    }                                                                                         │
│  }                                                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────╯
      r replay again  y copy  esc close
```

`r` from there re-sends the identical bytes and paints the identical panel. The hint bar offers nothing else, because `replayAgain` re-runs `m.replayReq` unchanged (`internal/tui/model.go:780`).

The fix is one string literal. Applying it means leaving mcpsnoop.

```sh
$ { printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"by-hand","version":"1"}}}'
    printf '%s\n' '{"jsonrpc":"2.0","method":"notifications/initialized"}'
    printf '%s\n' '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"search_orders","arguments":{"since":"2026-07-01","region":"emea","limit":5}}}'
    sleep 0.4; } | python3 srv.py
{"jsonrpc": "2.0", "id": 1, "result": {"protocolVersion": "2025-06-18", ...}}
{"jsonrpc": "2.0", "id": 2, "result": {"content": [{"type": "text", "text": "3 orders since 2026-07-01 in emea"}], "isError": false}}
```

That works, and mcpsnoop never sees it. Piping the same hand-written frames back through the shim does record them, but as a second session with its own id space and its own `clientInfo`, so comparing the two becomes a cross-session `mcpsnoop diff` instead of one overlay. The Inspector CLI is the packaged version of the same detour, `--tool-args-json '{"zip":"10001"}'` ([clients/cli/README.md](https://github.com/modelcontextprotocol/inspector/blob/main/clients/cli/README.md)), and it is a separate client that has never seen the capture.

Every proxy in this class draws the line the same way. Charles puts it in one sentence.

> The Edit tool builds on the Repeat tool by allowing you to change the contents of the request before repeating it.

([Charles, Edit tool](https://www.charlesproxy.com/documentation/tools/edit/).) mcpsnoop already has Repeat. It is missing Edit.

### One defect noticed while checking this

The overlay above printed `ok  (0s)` above a body carrying `"isError": true`, while the stream row for the very same response says `err`. `replayContent` branches only on the JSON-RPC error (`internal/tui/replay.go:42`), and `replay.Result` has no tool-error field (`internal/replay/replay.go:21`) the way `store.CallView.ToolErr` does (`internal/store/views.go:24`). It belongs here rather than in its own issue, because the entire value of an edit loop is telling at a glance whether the edit worked. If a reviewer prefers, it splits cleanly.

## Proposed change

Add `R` (shift+r) as *edit and replay*, everywhere `r` already replays, leaving `r` exactly as it is.

1. **Binding.** `R` is free where it is needed. In the stream view `applySortKey` does not claim it (`internal/tui/model.go:824-838`), verified by pressing it today and watching the view not move. In the sessions view `R` does sort by `req`, and that is fine because `startReplay` refuses anything that is not a focused request frame (`internal/tui/model.go:766`), so replay is never on offer there. Wire it beside the three existing `Replay` cases at `internal/tui/model.go:353`, `:357`, and `:421`.

2. **Edit surface.** `R` writes the captured `params`, pretty-printed, to a temp file and opens `$VISUAL` or `$EDITOR` through `tea.ExecProcess`, which releases the terminal and restores it afterwards. This also works for `mcpsnoop open -`, where the program's input is the controlling tty rather than the pipe carrying the log (`cmd/mcpsnoop/main.go:882-889`, `internal/tui/run.go:127`), and bubbletea hands the child that same reader.

3. **What gets edited.** The whole `params` object, not just `arguments`. Replay is method-general and takes a bare `params` (`internal/replay/replay.go:47`), so this covers `resources/read` and `prompts/get` for free.

4. **Validation.** The edited buffer has to parse as a JSON object. Anything else aborts with a flash and no request. This is not arbitrary strictness. `withClientMeta` already gives up and passes params through untouched when they are not an object (`internal/replay/replay.go:145`), so a non-object edit would behave one way against a legacy server and another against a stateless one.

5. **Send.** Pass the edited bytes to `replay.Replay` in place of the captured ones (`internal/tui/replay.go:24`). Store the edited params in `m.replayReq` so `r` from the result overlay iterates on the edit, and `R` from the result overlay re-opens the editor seeded with what was last sent.

6. **Display.** When sent and captured differ, the header reads `REPLAY · EDITED · tools/call` and the body shows both, `captured params` and `params sent`. `y` needs no change, it already copies the whole panel body (`internal/tui/model.go:1261-1264`).

7. **Safety.** The y/n gate that shows the recorded server command before running it stays exactly as it is (`internal/tui/model.go:796-809`). Editing never reaches a spawn that was not confirmed.

This stays off the live data path. The replayed server is a separate process spawned by mcpsnoop, and no replay result touches the store, so nothing an edit produces can be read back as observed traffic. That is also why this is not the breakpoint feature the same proxies ship, and why it does not collide with the rule that the proxy is transparent.

## Acceptance criteria

- [ ] `R` in the stream view, the inspector overlay, and the replay overlay opens the focused request's `params` in `$VISUAL` or `$EDITOR`, and `r` in all three still sends the captured bytes unchanged.
- [ ] The editor buffer is the pretty-printed `params` object. Saving it unmodified sends bytes equivalent to the capture.
- [ ] A buffer that does not parse, or parses to something other than a JSON object, aborts with a flash naming the parse failure. No request goes out and no model state is left half-updated.
- [ ] An empty buffer, or an editor exiting non-zero, aborts the same way.
- [ ] When the sent params differ from the captured ones, the overlay header is marked `EDITED` and the body shows `captured params` and `params sent` as separate blocks.
- [ ] `r` from an edited result re-runs the edited params. `R` from an edited result re-opens the editor holding what was last sent.
- [ ] The overlay reports a result carrying `"isError": true` as an error rather than `ok`, agreeing with the stream row for the same response. Today it does not, per the observed output above.
- [ ] The confirm-before-running-the-recorded-command prompt is unchanged, and no edit path can reach a spawn without it.
- [ ] No replayed frame, edited or not, enters the store, the JSONL log, `export`, `check`, `diff`, or `baseline`. This holds today and needs a test that keeps it holding.
- [ ] Against a stateless 2026-07-28 server, the params shown as sent are what actually went on the wire, including the `_meta` that `withClientMeta` merges in afterwards (`internal/replay/replay.go:112`).
- [ ] Tests drive a stub `$EDITOR` script covering accept-with-change, accept-unchanged, unparseable JSON, valid JSON that is not an object, and non-zero exit.
- [ ] The key table in `README.md:203` and the walkthrough in `docs/TRY_IT.md:48` mention the new key.

## Out of scope

- **No validation against the tool's `inputSchema`.** The store already holds it (`internal/store/views.go:438`, `ToolDefinition.InputSchema` at `internal/store/views.go:134`), so this is tempting, and it is still the wrong call here. The spec puts input validation on the server, under Security Considerations on the Tools page, "Servers **MUST**" then "Validate all tool inputs". A local pre-check would only add a second opinion that can disagree with the authority you are trying to observe, and being able to send a deliberately invalid argument is half the point of the feature. Worth a follow-up as an advisory hint if anyone wants it, never a gate.
- **No breakpoint and no editing on the live path.** This is a replay against a separately spawned server. The live client session is never touched.
- **HTTP sessions stay unreplayable.** That is #164. This inherits whatever shape that lands in.
- **MRTR continuations.** A captured `tools/call` carrying `requestState` is half of an exchange with a server that is now gone, and the spec says "Note that the JSON-RPC `id` **MUST** be different between the initial request and the retry". Editing and replaying such a request against a fresh server is not meaningful, and this issue does not try to make it so.
- **No edit history, no saving an edited request to a file, and no re-editing a past replay from a session opened later.** One live loop only.

## Files

- `internal/tui/model.go:353`, `:357`, `:421` for the three places `Replay` is dispatched, `:766` `startReplay`, `:780` `replayAgain`, `:787` `runReplay` and its confirm gate, `:824` `applySortKey` for the `R` collision check, `:1261` `copyCurrent`
- `internal/tui/keys.go:22` and `:48` for the new binding
- `internal/tui/replay.go:24` `replayCmd`, `:32` `replayContent` for the edited-versus-captured rendering and the `isError` header
- `internal/tui/view.go:1315` for the replay overlay hint bar, `:957` for the help screen line, `:186` and `:1308` for the stream and inspector hints
- `internal/replay/replay.go:21` `Result`, `:47` `Replay`, `:112` and `:145` `withClientMeta` for the stateless `_meta` merge
- `internal/store/views.go:15` `CallView`, `:134` `ToolDefinition.InputSchema`, `:611` `Command`
- `cmd/mcpsnoop/main.go:882` and `internal/tui/run.go:127` for the `open -` input reader
- `README.md:203`, `docs/TRY_IT.md:48`

shixi-li:
## What and why

Fixes #207.

`r` could replay a captured request, but fixing a bad argument required leaving mcpsnoop and reconstructing the call elsewhere. This adds `R` (shift+r) beside every replay entry point so users can edit the complete `params` object in `$VISUAL` or `$EDITOR`, validate it, and send it through the existing isolated replay path.

The implementation keeps three values distinct:

- the immutable captured params
- the user-edited replay candidate
- the actual wire params returned by `replay.Replay`, including stateless protocol `_meta`

Edited results show both captured and sent params, `r` repeats the current candidate, and `R` reopens it. Invalid, empty, non-object, or editor-failed buffers abort without changing replay state. The existing recorded-command confirmation still gates every spawn, temporary buffers are mode 0600 and removed on every callback path, and replay traffic remains outside the observed store and logs.

This also makes replay results classify MCP `result.isError: true` as an error instead of displaying `ok`.

## Checklist

- [x] Added or updated tests for the change, including editor failure cases and fail-before/pass-after behavior
- [x] Updated the README / docs for the new key
- [x] Windows cross-build passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ server is written in.
 |---|---|---|---|---|
 | `enter` | inspect / drill in | | `/` | filter |
 | `esc` | back | | `:` | command |
-| `j` / `k` | move | | `r` | replay a call |
+| `j` / `k` | move | | `r` / `R` | replay / edit and replay |
 | `g` / `G` | top / bottom | | `c` | capabilities |
 | `ctrl-f` / `ctrl-b` | page | | `s` | tool summary |
 | `p` | pause | | `y` | copy |
--- a/docs/TRY_IT.md
+++ b/docs/TRY_IT.md
@@ -58,7 +58,7 @@ well. When you're done, `mcpsnoop unwrap everything` puts it back.
    > run its long-running operation.
 
 4. Drill into a frame with `enter`, filter with `/`, inspect capabilities with
-   `c`, and replay a call with `r`.
+   `c`, replay a call with `r`, or edit its params and replay with `R`.
 
 Frames stream in as the client works. The quick calls come back OK, and the long
 one sends progress notifications and shows a visibly higher latency.
--- a/internal/replay/replay.go
+++ b/internal/replay/replay.go
@@ -139,6 +139,10 @@ func Replay(ctx context.Context, command []string, cwd, method string, params js
 	}, nil
 }
 
+// isToolError mirrors the store's own definition of a tool-level error, kept
+// here because the two packages are peers and neither imports the other. Change
+// one and check the other, or a replayed call and the captured call it came from
+// will disagree about the same bytes.
 func isToolError(result json.RawMessage) bool {
 	var r struct {
 		IsError bool `json:"isError"`
--- a/internal/replay/replay.go
+++ b/internal/replay/replay.go
@@ -25,6 +25,7 @@ type Result struct {
 	Response  json.RawMessage // raw response frame
 	RPCResult json.RawMessage
 	Err       *proxy.RPCError
+	ToolErr   bool // tools/call result.isError == true
 	Duration  time.Duration
 }
 
@@ -133,8 +134,16 @@ func Replay(ctx context.Context, command []string, cwd, method string, params js
 		Response:  resp,
 		RPCResult: msg.Result,
 		Err:       msg.Error,
+		ToolErr:   method == "tools/call" && isToolError(msg.Result),
 		Duration:  dur,
 	}, nil
+}
+
+func isToolError(result json.RawMessage) bool {
+	var r struct {
+		IsError bool `json:"isError"`
+	}
+	return json.Unmarshal(result, &r) == nil && r.IsError
 }
 
 // withClientMeta adds the self-describing _meta a stateless server expects,
--- a/internal/replay/replay_test.go
+++ b/internal/replay/replay_test.go
@@ -53,6 +53,8 @@ func handle(line []byte) {
 		if req.Params.Name == "echo" {
 			time.Sleep(120 * time.Millisecond)
 			out["result"] = map[string]any{"content": []map[string]string{{"type": "text", "text": "echo: " + req.Params.Arguments.Text}}}
+		} else if req.Params.Name == "tool-error" {
+			out["result"] = map[string]any{"content": []map[string]string{{"type": "text", "text": "bad argument"}}, "isError": true}
 		} else {
 			out["error"] = map[string]any{"code": -32601, "message": "unknown tool: " + req.Params.Name}
 		}
@@ -190,6 +192,20 @@ func TestReplayToolError(t *testing.T) {
 	}
 }
 
+func TestReplayToolResultError(t *testing.T) {
+	res, err := Replay(context.Background(), []string{demoBin}, "",
+		"tools/call", json.RawMessage(`{"name":"tool-error"}`), 10*time.Second)
+	if err != nil {
+		t.Fatalf("Replay transport error: %v", err)
+	}
+	if res.Err != nil {
+		t.Fatalf("tool-level error must not become a JSON-RPC error: %+v", res.Err)
+	}
+	if !res.ToolErr {
+		t.Fatalf("result.isError was not classified: %s", res.RPCResult)
+	}
+}
+
 func TestReplayEmptyCommand(t *testing.T) {
 	if _, err := Replay(context.Background(), nil, "", "tools/list", nil, time.Second); err == nil {
 		t.Fatal("expected error for empty command")
@@ -222,7 +238,7 @@ func TestReadResponseRejectsMalformedResponse(t *testing.T) {
 // the initialize error and carry on rather than giving up or hanging.
 func TestReplayAgainstStatelessServer(t *testing.T) {
 	res, err := Replay(context.Background(), []string{statelessBin}, "",
-		"tools/call", json.RawMessage(`{"name":"echo","arguments":{"text":"hi"}}`), 10*time.Second)
+		"tools/call", json.RawMessage(`{"name":"echo","arguments":{"text":"hi"},"_meta":{"progressToken":"p-7"}}`), 10*time.Second)
 	if err != nil {
 		t.Fatalf("Replay: %v", err)
 	}
@@ -233,6 +249,12 @@ func TestReplayAgainstStatelessServer(t *testing.T) {
 	// proves the request described itself.
 	if !strings.Contains(string(res.RPCResult), "stateless echo "+statelessProtocolVersion) {
 		t.Fatalf("result missing the self-describing metadata: %s", res.RPCResult)
+	}
+	if !strings.Contains(string(res.Params), statelessProtocolVersion) {
+		t.Fatalf("Result.Params must be the actual wire params with injected metadata: %s", res.Params)
+	}
+	if !strings.Contains(string(res.Params), `"progressToken":"p-7"`) {
+		t.Fatalf("injected metadata must preserve captured _meta values: %s", res.Params)
 	}
 }
 
--- a/internal/tui/keys.go
+++ b/internal/tui/keys.go
@@ -19,14 +19,15 @@ type keyMap struct {
 	Command key.Binding
 	Help    key.Binding
 
-	Replay  key.Binding
-	Caps    key.Binding
-	Summary key.Binding
-	Pause   key.Binding
-	Follow  key.Binding
-	Copy    key.Binding
-	Export  key.Binding
-	Delete  key.Binding
+	Replay     key.Binding
+	EditReplay key.Binding
+	Caps       key.Binding
+	Summary    key.Binding
+	Pause      key.Binding
+	Follow     key.Binding
+	Copy       key.Binding
+	Export     key.Binding
+	Delete     key.Binding
 
 	Quit key.Binding
 }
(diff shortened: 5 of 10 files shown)


kerlenton:
Great! Thanks!

PR #244, opened 2026-08-26
feat: expose live Prometheus tool metrics

### LINKED ISSUE #137: Expose tool call metrics on a Prometheus endpoint
## Context

Live OTLP export shipped, Prometheus never did. `--otlp-endpoint` streams completed calls as spans from the shim (`cmd/mcpsnoop/main.go:298`) and from `mcpsnoop http` (`cmd/mcpsnoop/main.go:767`), delivered by `internal/otlpsink`. Nothing exposes a scrape endpoint, no flag on either command mentions metrics, and the tree contains no `internal/metrics`.

The numbers exist, but not in the process a scrape would hit. `Store.ToolSummary` (`internal/store/views.go:675`) computes per-tool call counts, error counts, pending counts, nearest-rank P50, P95 and P99, and since #169 also per-tool result bytes and the largest single result. Tracing a three-call session through the shim and exporting it shows the whole set.

```
$ mcpsnoop export <session> --format json | jq .summary.tools
[
  { "name": "echo", "calls": 2, "errors": 1, "pending": 0,
    "p50_ms": 49.687, "p95_ms": 254.998, "p99_ms": 254.998,
    "result_bytes": 70, "max_result_bytes": 70 },
  { "name": "slow", "calls": 1, "errors": 0, "pending": 0,
    "p50_ms": 254.931, "p95_ms": 254.931, "p99_ms": 254.931,
    "result_bytes": 70, "max_result_bytes": 70 }
]
```

`ToolSummary` has exactly two callers, the TUI (`internal/tui/view.go:1402`) and the exporter (`internal/exporter/exporter.go:327`). `check` is not one of them; it builds a store through `exporter.LoadFile` but walks `st.Calls` itself for the latency budget (`cmd/mcpsnoop/check.go:166`), and a `check` run prints no per-tool block.

None of those is a place a scrape can reach. The TUI's store is fed live over the hub socket rather than read from a trace (`internal/tui/run.go:52`), and it ends when the user quits. The shim does already run the calculation, but only when `--otlp-endpoint` is set and only per completed call, on a store built and thrown away each time (`internal/otlpsink/sink.go:169`, through `exporter.Build`). Nothing anywhere holds a running aggregate, so serving a live endpoint is new accumulation rather than a second read of an existing one. The percentiles are the wrong shape as well, since `nearestRank` (`internal/store/views.go:753`) returns three exact quantiles from a retained duration slice while Prometheus wants bucket counters it can aggregate across scrapes.

## Decision to make before writing code

This is a step sideways from debugging in a terminal toward long-lived monitoring, which is where every competing tool went and where mcpsnoop deliberately did not. The local, no-account, single-binary position is the reason the tool has a niche at all. Worth deciding whether a metrics endpoint belongs in it before building one, rather than discovering the answer afterwards.

## Where it would live

The placement decides most of the design, and neither option is free.

- In the proxy. Every wrapped server is its own process. Running two wraps at once under one home produced two shims and two session logs, so a single fixed listen address collides and a per-shim address is awkward to write a scrape config against. `mcpsnoop http` also reverse-proxies every path to the target (`internal/proxy/http.go:297`), confirmed by a `GET /metrics` on the proxy listener coming back with the target server's own response, so the endpoint needs its own listener rather than a path on `--listen`.
- In the hub. `mcpsnoop` with no arguments already collects every shim into one store (`internal/tui/run.go:28`) and already calls `ToolSummary`, so this placement really is a second surface on an existing calculation. It is an interactive TUI that ends when the user quits, so scraping it needs a headless mode that does not exist yet.

## Proposed change

An opt-in flag taking an address, on its own listener, exposing per-tool call and error counters and a latency histogram. Write the text exposition format by hand rather than taking `client_golang`, because go.mod carries no telemetry dependency, OTLP is already hand-rolled in `internal/exporter` (`WriteOTLP`, `exporter.go:483`), and the README sells a single binary with no runtime deps. Follow Prometheus naming conventions and document the metric names, because they become an interface the moment anyone builds a dashboard on them. No cost when the flag is absent.

## Acceptance criteria

- [ ] Per-tool counters and a latency histogram are exposed
- [ ] Metric names follow convention and are documented
- [ ] Off by default, no overhead when disabled
- [ ] The endpoint has its own listener, so it cannot shadow a path `mcpsnoop http` forwards to the target
- [ ] Two concurrent shims do not fight over one address
- [ ] Tests cover emission

## Files

- new `internal/metrics`, wired in `cmd/mcpsnoop/main.go` beside the OTLP flags at `:298` and `:767`
- a proxy placement adds a `proxy.Sink` next to `otlpsink` in `traceSink` (`cmd/mcpsnoop/main.go:664`); a hub placement hooks the store in `internal/tui/run.go:28`
- the flag can stay CLI only, like `--otlp-endpoint`, which the `.mcpsnoop.toml` reader does not accept either (`cmd/mcpsnoop/config.go:100`)

bferanmi806-sketch:
Closes #137

Add an opt-in hub-side Prometheus listener. Bare mcpsnoop keeps its interactive TUI; mcpsnoop --metrics-listen runs the hub headlessly so the metrics process remains available without an interactive terminal.

Startup backfill primes session metadata and correlation state but is excluded from live counters. HTTP correlation includes ConnID, and MRTR operations remain one call with one latency observation. Reused request identities also clear stale pending tool state.

Public metrics: mcpsnoop_tool_calls_total, mcpsnoop_tool_errors_total, and mcpsnoop_tool_call_duration_seconds.

The listener is independent of the MCP proxy and uses stdlib-only Prometheus text exposition, with no new runtime dependency.

Tests run: targeted metrics, hub, TUI, and CLI tests; repeated regression tests for HTTP correlation, backfill priming, reconnect identity, headless lifecycle, and default routing; go vet ./...; staticcheck; cross-platform vet for Windows, Darwin, and Linux; git diff --check. The full go test ./... run is blocked on this Windows environment by existing Unix executable, file-mode, pipe, editor, and timestamp assumptions. POSIX make check could not run because the available POSIX shell cannot see go or gofmt. Race tests could not run because CGO/GCC is unavailable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,37 @@ since it names the auth scheme and the resource metadata to go to next. Filter b
 status with `status:401` in the TUI, or by any failure with `status:err`. A 4xx
 or 5xx counts as an error, so a default `mcpsnoop check` run fails on it.
 
+### Prometheus metrics
+
+Start the hub with an explicit metrics address to expose live tool-call metrics:
+
+```bash
+mcpsnoop --metrics-listen 127.0.0.1:9464
+curl http://127.0.0.1:9464/metrics
+```
+
+The listener is separate from the MCP proxy listener and is disabled unless
+`--metrics-listen` is provided. Startup history replay is not counted as new live
+traffic. Every series has `server`, `server_id`, and `tool` labels;
+`server_id` is a stable short fingerprint of the recorded server identity, so
+two servers with the same label remain separate without exposing commands,
+paths, endpoints, or session IDs. Error counters also have `error_type`, either
+`tool` for `result.isError` or `protocol` for a JSON-RPC or other protocol-level
+failure.
+
+The public metrics are:
+
+| Metric | Meaning |
+|---|---|
+| `mcpsnoop_tool_calls_total` | live tool-call requests observed |
+| `mcpsnoop_tool_errors_total` | live tool errors, split by `error_type` |
+| `mcpsnoop_tool_call_duration_seconds` | request-to-response latency histogram |
+
+The histogram exports the standard `_bucket`, `_sum`, and `_count` series with
+`0.005`, `0.01`, `0.025`, `0.05`, `0.1`, `0.25`, `0.5`, `1`, `2.5`, `5`, `10`,
+and `+Inf` second buckets. Pending, superseded, and calls cancelled without a
+result do not add latency observations.
+
 No server of your own? [Try it for real](docs/TRY_IT.md) against a published
 test server, driven by your own client. To inspect a session after it happened,
 see [review past sessions from logs](docs/POST_MORTEM.md).
@@ -146,6 +177,7 @@ Explicit command-line flags override values from the config file.
 |---|---|
 | `mcpsnoop -- <server>` | wrap a stdio server as a transparent shim |
 | `mcpsnoop` | open the live TUI |
+| `mcpsnoop --metrics-listen <addr>` | open the TUI and expose live Prometheus metrics |
 | `mcpsnoop http --target <url>` | proxy a streamable-HTTP server |
 | `mcpsnoop export` | render a session to json, html, text, har, or otlp |
 | `mcpsnoop check` | fail CI on errors, invalid frames, warnings, routing mismatches, hung calls, late results, or a latency budget |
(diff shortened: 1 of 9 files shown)


kerlenton:
LGTM! Thanks!

bferanmi806-sketch:
Thanks again for the review and merge! I genuinely enjoyed working on this.

If there are any other issues where an extra pair of hands would be useful, feel free to tag or assign me. I’d be very happy to help. And if there’s anything on the roadmap that you’d like someone to take on, I’d be glad to work on that too.

PR #245, opened 2026-09-10
fix(proxy): count events emitted after sink close

pacocartones:
## What and why

Count every event emitted after `AsyncSink.Close` as dropped. Previously the still-open buffered channel could accept the first late event even though no goroutine remained to consume it, so `Dropped()` underreported the loss.

A non-blocking read-lock attempt preserves the sink's no-backpressure contract during concurrent close, while the lifecycle flag makes post-close accounting exact. Regression tests cover the first late event and lock contention.

## Checklist

- [x] `make check` passes, including the race suite and cross-platform vet
- [x] Tests cover the behavior change
- [x] No documentation change is needed for this internal accounting correction


kerlenton:
@pacocartones LGTM! Thank you!

